Issue #14, opened 2022-01-10
add a flag to allow disable drift detection


chanwit:
Some resources might be very expensive to detect drift, for example a large cluster or complex Terraform resources.

It would be great to have a to disable drift detection.

`.spec.disableDriftDetection` would be suffice to support this feature.

chanwit:
@phoban01 would you like to try give it a try on this one please?

phoban01:
@chanwit Sure thing, I'll take this.

PR #20, opened 2022-01-12
Support disabling drift detection

phoban01:
This change introduces a new field `.spec.disableDriftDetection` which when `true` will bypass drift detection. Defaults to false.

This may be desirable when working with a complex state file or large Kubernetes cluster.

Example usage, disabling drift detection:
```yaml
kind: Terraform
metadata:
  name: helloworld
  namespace: flux-system
spec:
  approvePlan: "auto"
  disableDriftDetection: true
  path: ./
  sourceRef:
    kind: GitRepository
    name: helloworld
    namespace: flux-system
```

Implements #14 

Signed-off-by: Piaras Hoban <piaras@weave.works>

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,21 +3,21 @@
 `tf-controller` is an experimental controller for Flux to reconcile Terraform resources.
 
 ## Features
-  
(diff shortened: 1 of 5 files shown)


chanwit:
LGTM @phoban01 
Thank you so much 🥇

PR #21, opened 2022-01-12
Update README for auto approvePlan

tomhuang12:
Update README to reflect changes for automatic `approvePlan`

Signed-off-by: Tom Huang <tom.huang@weave.works>


chanwit:
Good catch, @tomhuang12 !
Thank you, and congrats for your first contribution!!

Issue #35, opened 2022-01-17
Run tests for the PRs


chanwit:

phoban01:
Going to take a look at this today.

PR #42, opened 2022-01-19
Test Suite Cleanup

phoban01:
This PR introduces the following changes in order to prepare the ground for test suite automation (#35) .

Items addressed are as follows:
- Increase global timeout to 30s to mitigate flaky asynchronous tests involving Terraform plan & apply operations
- Add environment variables to allow disabling Kubernetes and Terraform logs for more readable test output
- Update Makefile to use `go test -v ./controllers` so that test result output is streamed as the individual tests run (rather than at the end of the entire run).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ download-crd-deps:
 
 .PHONY: test
 test: manifests generate download-crd-deps fmt vet envtest ## Run tests.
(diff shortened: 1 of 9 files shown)


chanwit (approved):

chanwit:
Thank you so much @phoban01 :1st_place_medal: !!

PR #50, opened 2022-01-20
delete TF resources after each test

chanwit:


phoban01 (approved):
LGTM, thanks @chanwit . 

My only suggestion is that we could make a habit of using `defer` statements to declare the intention to clean up resources close to where the are created.

For example:
```golang
	By("creating the GitRepository resource in the cluster.")
	It("should be created successfully.")
	g.Expect(k8sClient.Create(ctx, &testRepo)).Should(Succeed())
	defer func() { g.Expect(k8sClient.Delete(ctx, &testRepo)).Should(Succeed()) }()
```

chanwit:
That would be a great idea. I'll change them to be defers.

phoban01 (approved):
LGTM, runs smooth. Thanks @chanwit :rocket:

PR #53, opened 2022-01-20
Add TCP and HTTP Health Checks

tomhuang12:
This is to add TCP/HTTP health checks.

```
# main.tf

output "rdsAddress" {
  value = "mydb.xyz.us-east-1.rds.amazonaws.com"
}

output "rdsPort" {
  value = "3306"
}

output "myappURL" {
  value = "https://example.com/"
}
```

```yaml
kind: Terraform
metadata:
  name: helloworld
  namespace: flux-system
spec:
  approvePlan: "auto"
  path: ./
  sourceRef:
    kind: GitRepository
    name: helloworld
    namespace: flux-system
  healthChecks:
    - name: rds
      type: tcp
      address: "{{.rdsAddress}}:{{.rdsPort}}" # uses standard Go package template format to parse outputs to url
      timeout: 10s # optional, defaults to 20s
    - name: myapp
      type: http
      url: "{{.myappURL}}"
      timeout: 5s
    - name: url_not_from_output
      type: http
      url: "https://example.org"
```

Discussions happening in https://github.com/chanwit/tf-controller/issues/44


tomhuang12:
@chanwit This is ready for review. Thanks!

README.md
```diff
@@ -305,6 +305,53 @@ kubectl create secret \
...
+  approvePlan: "auto"
+  path: ./
+  sourceRef:
+    kind: GitRepository
+    name: helloworld
+    namespace: flux-system
+  healthChecks:
+    - name: rds
+      type: tcp
+      url: "{{.rdsUrl}}:{{.rdsPort}}" # uses standard Go package template format to parse outputs to url
```
chanwit, line 344: The concept of using `url` is better than `address` 👍 

To make it consistent, `url` would be specified only in one of the following forms:
- `tcp://{{.rdsUrl}}:{{.rdsPort}}`
- `http://...`
- `https://...`
maybe we need a check for this pattern.

api/v1alpha1/reference_types.go
```diff
@@ -58,3 +61,45 @@ type VarsReference struct {
...
+	Name string `json:"name"`
+
+	// Type of the health check, valid values are ('tcp', 'http').
+	// +kubebuilder:validation:Enum=tcp;http
+	// +required
+	Type string `json:"type"`
+
+	// URL to perform the health check on.
+	// Go template can be used to reference values from the terraform output
+	// (e.g. url: "{{.foo}}:{{.bar}}").
```
chanwit, line 81: e.g. `url: "tcp://{{.managed_service_ip}}:{{.managed_service_port}}"`

controllers/tc000240_health_check_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	tcpTestCases := []struct {
+		name    string
+		url     string
+		timeout time.Duration
+		wantErr bool
+	}{
+		{
+			name:    "testTCP",
+			url:     "weave.works:80",
```
chanwit, line 27: Please check it to:
`url: "tcp://weave.works:80",`
and specifying only `url: "weave.works:80",` would fail because the string is not in the form of URL.

controllers/terraform_controller.go
```diff
@@ -586,9 +607,11 @@ terraform {
 
 	}
 
+	justApplied := false
 	outputs := map[string]tfexec.OutputMeta{}
 	if r.shouldApply(terraform) {
 		terraform, err = r.apply(ctx, terraform, tf, revision, &outputs)
+		justApplied = true
```
chanwit, line 614: This `justApplied` check is not enough here.
Please check that the Apply condition is "Success" before doing health checks.

controllers/terraform_controller.go
```diff
@@ -1509,3 +1544,137 @@ func (r *TerraformReconciler) gzipDecode(encodedPlan []byte) ([]byte, error) {
...
+					err.Error(),
+				), err
+			}
+			parsedURL = b.String()
+		}
+
+		// perform health check based on type
+		switch hc.Type {
+		case infrav1.HealthCheckTypeTCP:
+			err := r.doTCPHealthCheck(ctx, hc.Name, parsedURL, hc.GetTimeout())
```
chanwit, line 1600: Could we make sure that the parsedURL is in the form of `tcp://...` here?

controllers/terraform_controller.go
```diff
@@ -1509,3 +1544,137 @@ func (r *TerraformReconciler) gzipDecode(encodedPlan []byte) ([]byte, error) {
...
+		}
+	}
+	terraform = infrav1.TerraformHealthCheckSucceeded(terraform, "Health checks succeeded")
+	return terraform, nil
+}
+
+func (r *TerraformReconciler) doTCPHealthCheck(ctx context.Context, name string, url string, timeout time.Duration) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	conn, err := net.DialTimeout("tcp", url, timeout)
```
chanwit, line 1632: this `url` variable is a bit miss-leading because it's an address, not a url.

chanwit (changes requested):
Thank you so much @tomhuang12 
This feature is a huge plus. It makes the managed service use case very practical !!

There are some changes required to make the semantic of `url` being more consistent, and 
to improve the trigger logic of the health check process.

tomhuang12:
@chanwit Thanks for the review! In order to make `tcp://blah:8080` format to work, I would need to drop `tcp://` part and just use the address. Is that what you like? Reference: https://pkg.go.dev/net#Dial

chanwit:
To make the UX semantically consistent, we should consider either this, having `address` field for TCP and `url` for HTTP:
````
  healthChecks:
    - name: rds
      type: tcp
      address: "1.1.1.1:3000" 
    - name: myapp
      type: http
      url: "https://a.b.c"
```

or this (the design we're discussing), having url with `tcp://` for TCP and `http://` for HTTP.
````
  healthChecks:
    - name: rds
      type: tcp
      url: "tcp://1.1.1.1:3000" 
    - name: myapp
      type: http
      url: "https://a.b.c"
```

which one you think it's better?

tomhuang12:
I like the first one better. I will make the changes. Thanks!

chanwit (approved):

chanwit:
LGTM 
thank you @tomhuang12 :1st_place_medal: !!

PR #66, opened 2022-01-28
Support HCL types in variables

### LINKED ISSUE #61: Add flag to specify that a variable is an HCL expression
To help with a correct variable generation, we need a flag to specify that a variable is an HCL expression. It's a string by default.

phoban01:
### What does this PR change

- Enables support for HCL variable types such as string, bool, number, object, list etc... on the Terraform custom resource.
- Updates outputs so that the JSON value is written to the output secret without the Terraform wrapper fields.
- Refactor generate vars code into dedicated file for code management 
- New spec and fixtures to cover HCL vars 

The primary change here is  from type `string` to `apiextensionsv1.JSON ` on the `TerraformSpec.Vars.Variable.Value` field.

### Example
The following HCL variable:

```hcl
variable "cluster_spec" {
  type = object({
      region = string
      env = string
      node_count = number
      public = bool
  })
}
```
Can be supplied thus:
```yaml
kind: Terraform
metadata:
  name: helloworld
  namespace: flux-system
spec:
  approvePlan: "auto"
  path: ./
  sourceRef:
    kind: GitRepository
    name: helloworld
    namespace: flux-system
  vars:
  - name: cluster_spec
    value:
      region: us-east-1
      env: dev
      node_count: 10
      public: false

```

Resolves #61


README.md
```diff
@@ -263,6 +263,40 @@ spec:
...
+  namespace: flux-system
+spec:
+  approvePlan: "auto"
+  path: ./
+  sourceRef:
+    kind: GitRepository
+    name: helloworld
+    namespace: flux-system
+  vars:
+    cluster_spec:
```
chanwit, line 293: From reading the codes,  I think this example should be?
```
spec:
  approvePlan: "auto"
  path: ./
  sourceRef:
    kind: GitRepository
    name: helloworld
    namespace: flux-system
  vars:
  - name: cluster_sepc
    value:
      region: us-east-1
      env: dev
      node_count: 10
      public: false    
```

chanwit (changes requested):
Thank you @phoban01 
I didn't think it became a powerful feature like this.

I really like the way of how you implemented this PR.
Thank you so much.

Small glitches around the document part.
Other than that it LGTM.

README.md
phoban01: Yup, thanks for catching @chanwit.

phoban01: Looks like the older example was wrong also; both now fixed.

chanwit:
Thank you very much @phoban01 :1st_place_medal: !!

Issue #80, opened 2022-02-03
`.spec.vars` is not working properly


tomhuang12:
Tried setting `vars` to the followings:

```
  vars:
  - name: region
    value: us-east-1
```
and 
```
vars:
  - name: cluster_spec
    value:
      region: us-east-1
      env: dev
      node_count: 10
      public: false
```

Both resulted the following error:
```
E0203 04:57:12.795883       7 reflector.go:138] pkg/mod/k8s.io/client-go@v0.23.1/tools/cache/reflector.go:167: Failed to watch *v1alpha1.Terraform: failed to list *v1alpha1.Terraform: json: cannot unmarshal object into Go struct field Variable.items.spec.vars.value of type string
```

phoban01:
Thanks for catching @tomhuang12. This issue was caused by the wrong container image being specified. 

I've pushed a fix in #82.

Issue #78, opened 2022-02-02
Test 230 and 140 fail consistently


chanwit:
Just the only test case I've seen failing repeatedly across job runs. Need a proper way to fix it.

tomhuang12:
Test 140 has the same issue.

chanwit:
updated

phoban01:
I believe this is happening because we have a variable shadowing problem in the drift detection block:
https://github.com/weaveworks/tf-controller/blob/eb1bd17016516d061657adf5a0a63e84eeff8de1/controllers/terraform_controller.go#L582-L606

If we want to keep these variables names it should read:
```golang
if r.shouldDetectDrift(terraform, revision) {
	var driftDetectionErr error // declared here to avoid shadowing on terraform variable
	terraform, driftDetectionErr = r.detectDrift(ctx, terraform, tfInstance, runnerClient, revision)
...
``` 
If we passed a pointer to the `detectDrift` method then we would also avoid this issue; that would involve changing the code in quite a few areas but might be worth it. 

Any thoughts @chanwit?

chanwit:
This is making sense because it's happening only with tests related to the drift detection.

PR #81, opened 2022-02-03
Add e2e testing

tomhuang12:
This is replacing regular test on PR to a more comprehensive e2e test. I am following the e2e test pattern of kustomize-controller. It has the following functionalities:
- Cache go modules
- Cache docker layers
- Set up v1.21.1 kind cluster for testing
- Runs through the most common use cases

Comments:
- Example successful run: https://github.com/tomhuang12/tf-controller/actions/runs/1787759833 (This run had `make test` commented out to by pass transient error relating to drift detection kept happening. https://github.com/weaveworks/tf-controller/issues/78 is tracking the error.)
- While writing test case, I found that `.spec.vars` isn't working properly like what's written in the README so I created https://github.com/weaveworks/tf-controller/issues/80 to track that. `config/testdata/vars/test.yaml` needs to be updated when the issue is addressed. 

Fix https://github.com/weaveworks/tf-controller/issues/51


config/manager/manager.yaml
```diff
@@ -3,7 +3,6 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: tf-controller
-  namespace: flux-system
```
tomhuang12, line 6: I moved `namespace` to the `kustomization.yaml` so that `make deploy` can install tf-controller in `tf-system` namespace while still keeping the deployment release manifest to have `flux-system` as its namespace.

phoban01:
Thanks @tomhuang12; this is very much needed and is helping surface many things already.

chanwit:
@tomhuang12 I think there's a test failure somehow.

chanwit (approved):
LGTM

Thank you @tomhuang12

PR #82, opened 2022-02-03
Image Registry Update

phoban01:
**What does this PR do**

This PR updates our Makefile and GitHub actions workflows to use the `ghcr.io/weaveworks` container registry.

PR #83, opened 2022-02-06
move import path and image to weaveworks

### LINKED ISSUE #78: Test 230 and 140 fail consistently

### LINKED ISSUE #79: Rename packages to be weaveworks/tf-controller
Need to fix package in many places to make sure the package name's changed to `weaveworks/tf-controller`.

chanwit:
fix #79
in addition to #82 

also, fix #78

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,7 +27,7 @@ RUN unzip -q /terraform_1.1.4_linux_amd64.zip
 
 FROM alpine:3.15
 
-LABEL org.opencontainers.image.source="https://github.com/chanwit/tf-controller"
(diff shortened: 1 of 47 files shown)


chanwit:
@phoban01 I think unshadowing the variable like you suggested did the trick.

phoban01 (approved):
LGTM, thanks @chanwit.

phoban01:
> @phoban01 I think unshadowing the variable like you suggested did the trick.

I think using a pointer will be the right fix long term here, as that's really the behavior we're after in this case. But it's better to do that refactor in it's own PR I think :smile:

PR #88, opened 2022-02-07
Enable mTLS

phoban01:
**What does this PR do:**

Enables support for mTLS connections between the Terraform controller and runner pods by adding an `mtls/` package containing a cert rotator which is  based on that used in Gatekeeper.

### DIFF
--- a/controllers/terraform_controller.go
+++ b/controllers/terraform_controller.go
@@ -43,6 +43,7 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 	"github.com/hashicorp/terraform-exec/tfexec"
 	infrav1 "github.com/weaveworks/tf-controller/api/v1alpha1"
+	"github.com/weaveworks/tf-controller/mtls"
 	"github.com/weaveworks/tf-controller/runner"
 	"github.com/zclconf/go-cty/cty"
 	ctyjson "github.com/zclconf/go-cty/cty/json"
@@ -55,6 +56,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	kuberecorder "k8s.io/client-go/tools/record"
 	"k8s.io/client-go/tools/reference"
 	"sigs.k8s.io/cli-utils/pkg/kstatus/polling"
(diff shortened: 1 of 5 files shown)


controllers/terraform_controller.go
```diff
@@ -1605,3 +1583,137 @@ func (r *TerraformReconciler) templateParse(content map[string]string, text stri
...
+			Namespace: terraform.Namespace,
+			Name:      fmt.Sprintf("%s-runner", terraform.Name),
+		},
+	}
+
+	if err := r.Get(ctx, client.ObjectKeyFromObject(runnerSvc), runnerSvc); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return err
+		}
+		runnerSvc.Spec = corev1.ServiceSpec{
```
chanwit, line 1652: @phoban01 Could you avoid using Service objects, and trying to connect directly to the runner pod via the pod FQDN first, please? Having a Service defined for every runner Pod will be in the plan B, if we cannot make use of Pod FQDN.


The FQDN for a pod could be obtained by prefixing its .status.PodIP with namespace and .pod.cluster.local (or the cluster domain suffix)

Here's the doc:
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods

phoban01: Hmm, @chanwit  thinking about this a bit more: we want mTLS to secure pod-pod communication in which case we need either a pod-ip or pod-hostname to issue the cert (otherwise we end up sharing certs). But the pod-ip isn't assigned until the pod is scheduled at which point it is too late to issue the cert (as it needs to be mounted in the pod at startup time). 

We have two options as far as I can see:
1) Create a headless service in the namespace, and specify the hostname and subdomain in the pod spec. This will allow us to connect to pods using the FQDN of the pod which will be `{hostname}.{subbomain}.{namespace}.svc.cluster.local`
2) Create one service per pod

I don't believe this would be any different using admission control as the pod-ip would not be assigned at that point either. Maybe there's another approach?

chanwit: This is an interesting problem. Could a SAN wildcard `*.{namespace}.pod.cluster.local` solve this?

phoban01: Yes that would work, with the proviso that mTLS is scoped at the namespace level rather than per-pod.

chanwit: That's great. Thank you, @phoban01.

phoban01: Have made a change to address this @chanwit.

phoban01:
I'll add some test coverage to verify the certificate generation process and I think we can then look at merging this. We can address the rotation in a separate PR, so that work can continue in parallel on the runner pod invocation. Sound good @chanwit @tomhuang12 ?

tomhuang12:
Sounds good! Thank you so much!

chanwit:
The test was still falling because Test 250 was trying to obtain the TLS secret, but panicked.

https://github.com/weaveworks/tf-controller/runs/5114747321?check_suite_focus=true#step:8:760

phoban01:
Hmm, I'm seeing a lot of tests failing with connection refused errors.

phoban01:
Added connection close handler to the LookupOrCreate runner reconciler method which will help with those connection errors.

PR #101, opened 2022-02-15
implement tf-runner pod and main program

### LINKED ISSUE #70: Implement runner main program
Implement runner main program and package it into another container image.

chanwit:
This PR implements TF-runner pod invocation, and the main runner program.

Fix #70 #71 and #73

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -1,0 +1,38 @@
+name: test
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+
+permissions:
+  contents: read # for actions/checkout to fetch code
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Restore Go cache
+        uses: actions/cache@v1
+        with:
+          path: ~/go/pkg/mod
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Cache Docker layers
+        uses: actions/cache@v2
+        id: cache
+        with:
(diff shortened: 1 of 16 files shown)


.github/workflows/test.yaml
```diff
@@ -0,0 +1,38 @@
+name: test
```
tomhuang12, line 1: The e2e action also does a `make test`. Do we need a separate action to do it again?

controllers/terraform_controller.go
```diff
@@ -1597,74 +1620,106 @@ func (r *TerraformReconciler) reconcileRunnerSecret(ctx context.Context, terrafo
...
+		time.Sleep(2 * time.Second)
+	}
+
 }
 
-func runnerPodSpec(tlsSecretName string) corev1.PodSpec {
+func getRunnerPodImage() string {
+	runnerPodImage := os.Getenv("RUNNER_POD_IMAGE")
+	if runnerPodImage == "" {
+		runnerPodImage = "ghcr.io/weaveworks/tf-runner:latest"
```
tomhuang12, line 1671: Should we pin to an image tag that matches the controller image tag?

.github/workflows/test.yaml
chanwit: yes. It's intended. Running them in parallel gives feedback faster.

controllers/terraform_controller.go
chanwit: We can pin it via the env here for every release: https://github.com/weaveworks/tf-controller/pull/101/files#diff-ce0d597a9e5f20a8fe61c3ba5e6185a6b90cfef2c00bb5b5ef3e97b15b79f480R49

Making it as an env so that it's flexible during the development and test.

Also, it allows hotfix when something unexpected breaks the users.

tomhuang12 (approved):
LGTM! 👍

chanwit:
yay!

PR #112, opened 2022-02-16
Exponentially increase wait time for runner pod

### LINKED ISSUE #107: Use exponential backoff to wait for runner pods instead of the current linear loop

tomhuang12:
Fix #107

### DIFF
--- a/controllers/terraform_controller.go
+++ b/controllers/terraform_controller.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"html/template"
 	"io"
+	"math/rand"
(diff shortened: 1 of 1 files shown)


don4of4:
@mmubeen have you seen this pr by @tomhuang12 ?  It looks like experimental backoff, which I thought wasn't part of the solution. CC @eparisca Regardless I think we need to open a PR around that fact that the statefiles are being saved every failure, which is exploding our blobs.

don4of4:
I see you created the feature here #1335.

PR #113, opened 2022-02-16
Rotate cert by leader only

### LINKED ISSUE #102: Only the leader generates / rotates the mTLS secret
Currently we don't have a semantic to check if the current instance of controller is the leader.
We have to make sure that only the leader (via lease) is allowed to generate and rotate the mTLS secret.

chanwit:
This PR use `<- mgr.Elected()` to block the cert rotation on other instances, so that the leader does so. 
Please note that `<-mgr.Elected()` gets closed only when the pod becomes the leader.

At the controller, instead of using the Ready ch, we read and validate the Cert secret directly. If it's not found, or invalid, we retry.

Fix #102


tomhuang12 (approved):
LGTM!

Issue #108, opened 2022-02-15
Implement more e2e tests to cover multi-tenancy setup


chanwit:

tomhuang12:
I will work on this when #106 is ready 😄

chanwit:
Please set controller's `replicas` to be 3 this time :-)

chanwit:
@tomhuang12 please see https://github.com/weaveworks/tf-controller/tree/main/config/testdata/multi-tenancy as the starting point. I put a simple one there.

Issue #104, opened 2022-02-15
Add an option to delete the runner pod for every reconciliation loop


chanwit:
For better security, we should support an option to delete the runner pod of a Terraform object when each loop is completed.

The new pod should be re-created when the new loop starts.

chanwit:
With this flag,
```
spec:
  alwaysCleanupRunner: true
```
the controller always delete the runner pod after use, so that we can make sure the pod is not left unintentionally.

PR #117, opened 2022-02-17
Improving e2e test

tomhuang12:
- Added a more complex multi-tenant test case #108 
- Added simple chaos testing by randomly deleting runner pod #115 
- Added some labels to the runner pod


tomhuang12:
Something I observed that the runner pod stays alive indefinitely for each `Terraform` object. If the user has a lot of `Terraforms`, then there will be a lot of long-living pods. Anything we should do to address this?

chanwit:
> Something I observed that the runner pod stays alive indefinitely for each Terraform object. If the user has a lot of Terraforms, then there will be a lot of long-living pods. Anything we should do to address this?

Yes, we'll address it by #104

chanwit (approved):
LGTM thank you @tomhuang12 :1st_place_medal:

PR #121, opened 2022-02-18
add always cleanup pod to the api

### LINKED ISSUE #104: Add an option to delete the runner pod for every reconciliation loop

chanwit:
Fix #104

### DIFF
--- a/controllers/terraform_controller.go
+++ b/controllers/terraform_controller.go
@@ -146,7 +146,6 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if closeConn != nil {
 			if err := closeConn(); err != nil {
 				log.Error(err, "unable to close connection")
-				retResult = ctrl.Result{}
 				retErr = err
 			}
 		}
@@ -162,7 +161,6 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			if err := cli.Get(ctx, podKey, &pod); err != nil {
 				if !apierrors.IsNotFound(err) {
 					log.Error(err, "unable to get pod for cleaning up")
-					retResult = ctrl.Result{}
 					retErr = err
 				}
 				return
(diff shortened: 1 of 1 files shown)


phoban01 (changes requested):
Can we use named return variables to capture any errors from defer'd functions? Looks good to me but my only concern is that we could be failing silently in some of our defer'd cleanup functions.  Trivial example:
```
func convert(x string) (result int, retErr error) {
	defer func() {
		result, retErr = strconv.Atoi(x)
	}()
	return 0, nil
}
```

controllers/terraform_controller.go
```diff
@@ -146,6 +146,8 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if closeConn != nil {
 			if err := closeConn(); err != nil {
 				log.Error(err, "unable to close connection")
+				retResult = ctrl.Result{}
+				retErr = err
```
phoban01, line 150: I think we only need to worry about capturing the error here.

controllers/terraform_controller.go
```diff
@@ -361,6 +397,113 @@ func (l LocalPrintfer) Printf(format string, v ...interface{}) {
...
+	terraform, tfInstance, tmpDir, err := r.setupTerraform(ctx, runnerClient, terraform, sourceObj, revision, objectKey)
+
+	defer func() {
+		cleanupDirReply, err := runnerClient.CleanupDir(ctx, &runner.CleanupDirRequest{TmpDir: tmpDir})
+		if err != nil {
+			log.Error(err, "clean up error")
+		}
+		if cleanupDirReply != nil {
+			log.Info(fmt.Sprintf("clean up dir: %s", cleanupDirReply.Message))
+		}
```
phoban01, line 414: These errors will be swallowed.

phoban01 (approved):
A few minor comments but LGTM; thanks @chanwit!

PR #123, opened 2022-02-18
remove the terraform binary from the controller image

### LINKED ISSUE #109: Remove Terraform binary from the tf-controller image Dockerfile (and not tf-runner image)

chanwit:
We no longer need the Terraform binary inside the controller image because it's now in the Runner Pod image.

Fix #109


tomhuang12 (approved):
LGTM

PR #129, opened 2022-02-21
implement output name mapping

### LINKED ISSUE #128: Add a test case for writing output with dots in its name to a secret
Need a test case to make sure that outputs containing dots, like the following, in their names are allowed:

```yaml
kind: Terraform
metadata:
  name: master-key-tf
  namespace: app-01
spec:
  interval: 1h
  path: ./_artifacts/10-zz-terraform
  writeOutputsToSecret:
    name: age
    outputs:
    - age.agekey
```

chanwit:
This PR allows output name mapping so that we can map an output name from Terraform to be something, like a key name, expected by a Flux controller.

For example, `age_key` can be mapping to `age.agekey` to be consumed by the Kustomization controller.

Fix #128


phoban01 (commented):
Looking at how this functionality is developing it makes me wonder if the API should be something like the following for outputs:

```yaml
writeOutputs:
   - kind: Secret
     name: helloworld-output
     outputs:
     - key: age_key
       targetPath: age.agkey
```

phoban01 (approved):
LGTM @chanwit

chanwit:
It would, yes. I agree with your cleaner suggestion. We can bring it to `v1alpha2`.

PR #144, opened 2022-03-03
Make cert rotation configurable

### LINKED ISSUE #125: Make cert rotation period configurable and add E2E test to check the behavior of Pods when cert is expired
Currently, we still don't have a set of E2E tests to ensure the behavior of controller + pods when the mTLS cert is expired.
To do so,
1. We make the cert rotation period configurable
2. Make the value small
3. Write test case to expect that the controller generate a new cert and install it for new pods. Existing pods would be unable to connect from the controllers. Some error would be expected, and the pods would be deleted.
4. Test 3 controllers, and the leader would be only the instance to do no 3.

tomhuang12:
This PR adds the ability to specify `--cert-validity-duration` and `--cert-rotation-check-frequency` flags on the controller to change the mTLS certificate validity duration and check frequency. 

Fix #125

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -63,16 +63,18 @@ func init() {
 
 func main() {
 	var (
-		metricsAddr           string
-		eventsAddr            string
-		healthAddr            string
-		concurrent            int
-		requeueDependency     time.Duration
-		clientOptions         client.Options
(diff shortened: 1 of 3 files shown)


chanwit:
LGTM thank you @tomhuang12 
tested it locally and it went through.

PR #146, opened 2022-03-08
use retry policy for grpc connection

chanwit:
This PR gives a block of 10 secs before gRPC going connection timed out.


phoban01 (commented):
@chanwit You could pass a retry policy `grpc.WithDefaultServiceConfig` to `grpc.Dial` rather than using context timeout. Full example is here: grpc.WithDefaultServiceConfig

What do you think?

chanwit:
That's super cool! I'll re-write it.

phoban01 (approved):
LGTM, nice work @chanwit

chanwit:
seems we need to rewrite whole E2E tests in Golang.

PR #156, opened 2022-03-11
use tlscert instead of cacert when validating cert of each pod

### LINKED ISSUE #155: TLSCert should be used to validate the host instead of CACert
Currently CACert is being used to validate the host instead of TLSCert. So the connection is failing in RC4 and RC5.

https://github.com/weaveworks/tf-controller/blob/091665b22a363bba6479b551a13a1cfa9f5a92d5/controllers/terraform_controller.go#L1813

We also need a test to prevent future issues of this kind.

chanwit:
Fixes #155 

This PR fixes the validation process of tlscert.

I removed batch pod deletion logic because it might accidentally delete long running pod of other TF objects.
We should let a TF object takes care of its own runner pod only.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -175,11 +175,11 @@ jobs:
       - name: Verify chaos testing result
         run: |
           kubectl -n chaos-testing get pods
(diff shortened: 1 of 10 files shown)


chanwit:
all green in a while!

phoban01 (approved):
LGTM, thanks @chanwit

PR #159, opened 2022-03-14
fix drifted outputs

### LINKED ISSUE #158: Even if plan is no changed, TF outputs must be reconciled if they're deleted
There is a corner case that TF outputs of a Terraform object get deleted, and no drift occurred, no plan is changed.
TF-controller should reconcile the TF outputs as a secret too so that other objects that expected this output could work normally.

chanwit:
Fixes #158 

This PR fixes drifted outputs and add an interval to the tests.

### DIFF
--- a/controllers/terraform_controller.go
+++ b/controllers/terraform_controller.go
@@ -434,6 +434,16 @@ func (r *TerraformReconciler) reconcile(ctx context.Context, runnerClient runner
 
 		// immediately return if no drift - reconciliation will retry normally
 		if driftDetectionErr == nil {
+			// reconcile outputs only when outputs are missing
+			if outputsDrifted, err := r.outputsMayBeDrifted(ctx, terraform); outputsDrifted == true && err == nil {
+				terraform, err = r.processOutputs(ctx, runnerClient, terraform, tfInstance, revision)
+				if err != nil {
+					return terraform, err
+				}
+			} else if err != nil {
+				return terraform, err
+			}
+
 			return terraform, nil
 		}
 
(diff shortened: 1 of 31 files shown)


controllers/terraform_controller.go
```diff
@@ -487,33 +496,82 @@ func (r *TerraformReconciler) reconcile(ctx context.Context, runnerClient runner
...
+
+	log := ctrl.LoggerFrom(ctx)
+	objectKey := types.NamespacedName{Namespace: terraform.Namespace, Name: terraform.Name}
+
+	outputs := map[string]tfexec.OutputMeta{}
+	var err error
+	terraform, err = r.obtainOutputs(ctx, terraform, tfInstance, runnerClient, revision, &outputs)
+	if err != nil {
+		return terraform, err
+	}
```
phoban01, line 529: It would make more sense to create the outputs map inside `obtainOutputs` and return it from the from. Likewise, I would also pass the `terraform` struct as a pointer rather than passing and returning it.

phoban01 (changes requested):
Thanks for adding chaos tests @chanwit; that will be really helpful for us all 😄

controllers/terraform_controller.go
chanwit: nop not me.
chaos tests has been added by @tomhuang12 a while ago. I just make it less CPU-consuming, but it's still meaningful enough for e2e.

chanwit: I did think that as well.

But this pattern is generally suggested by k8sclient, controller SDK, as well as KS controller. so I think we can just leave it there to conform to the overall writing style.

phoban01 (approved):

PR #172, opened 2022-03-19
add concurrency to the chart

### LINKED ISSUE #171: Add a helm chart value for specifying concurrency

chanwit:
Fixes #171


charts/tf-controller/templates/deployment.yaml
```diff
@@ -25,10 +25,10 @@ spec:
       containers:
       - args:
         - --watch-all-namespaces
-        - --log-level=info
+        - --log-level={{ .Values.logLevel }}
         - --log-encoding=json
         - --enable-leader-election
-        - --concurrent=1
+        - --concurrent={{ .Values.concurrency | toString }}
```
tomhuang12, line 31: I think it should be {{ .Values.concurrency | quote }} if you want to make it a quoted string.

PR #198, opened 2022-04-04
Setup a SecurityGroupPolicy

Nalum:
This change will allow setting up an [AWS EKS `SecurityGroupPolicy`](https://docs.aws.amazon.com/eks/latest/userguide/security-groups-for-pods.html#deploy-securitygrouppolicy) against the `ServiceAccount` used by the `tf-controller` `POD`s.

If supplied the `POD`s will be started and run under the specific AWS Security Groups that are passed in the list.

### DIFF
--- a/charts/tf-controller/Chart.yaml
+++ b/charts/tf-controller/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: tf-controller
 description: The Helm chart for Weave GitOps Terraform Controller
 type: application
-version: 0.2.5
+version: 0.2.6
 appVersion: "v0.9.3"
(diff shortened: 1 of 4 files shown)


phoban01:
Do we have an issue describing why this is required?

phoban01:
@Nalum Seems a bit specific to AWS; but would be good to have some context in the description or in an issue 🙂 .

Nalum:
@phoban01 I've updated the description, will think a bit more about info that I can add to it.

As to being specific to AWS, it is very much so.

tomhuang12 (approved):
LGTM! Thank you!

PR #199, opened 2022-04-04
Fix JSON encoding bug

phoban01:
**What does this PR do**
This PR addresses an issue where newlines or other characters passed as Terraform variables were not encoded correctly.

This PR change how we handle JSON encoding for a slice of bytes. Without first converting to a string or other type, a slice of bytes will be JSON encoded as a base64 string. Previously we had worked around this using `fmt.Sprintf` however this produced inconsistencies in the encoding. Here we instead convert the slice of bytes to a string which we then encode using `json.Marshal`.


controllers/tc000121_destroy_on_delete_test.go
```diff
@@ -105,15 +105,15 @@ func Test_000121_destroy_on_delete_test(t *testing.T) {
 			Vars: []infrav1.Variable{
 				{
 					Name:  "kubeconfig",
-					Value: utils.JsonEncodeBytes([]byte(testEnvKubeConfigPath)),
+					Value: utils.MustJSONEncodeBytes([]byte(testEnvKubeConfigPath)),
```
bigkevmcd, line 108: I don't think the `Must` idiom is appropriate here?

The string that is being encoded is provided by users, and so, it could be not-encodable as valid JSON, `panic()` can cause the process to crash, which doesn't sound like a desirable outcome?

chanwit:
LGTM, thank you @phoban01

PR #201, opened 2022-04-07
Use imagePullSecrets on ServiceAccounts

### LINKED ISSUE #200: Image Pull Secrets on ServiceAccounts
I have deployed the `tf-controller` on a cluster using the Helm Chart and defined the `imagePullSecrets` as we need to pull the controller and runner containers from a private registry.

The controller deploys without issue using the defined `imagePullSecrets` but the runner fails because it doesn't have the `imagePullSecrets` defined for the pod.

Given that there is a `ServiceAccount` definition for the controller and the runner I think it would make sense to have the `imagePullSecerts` applied to these rather than to the controller `Deployment`, or perhaps allow defining `imagePullSecrets` against both `ServiceAccount`s as well as the `Deployment`.

Nalum:
Closes #200

### DIFF
--- a/charts/tf-controller/Chart.yaml
+++ b/charts/tf-controller/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: tf-controller
 description: The Helm chart for Weave GitOps Terraform Controller
 type: application
-version: 0.2.6
+version: 0.2.7
 appVersion: "v0.9.3"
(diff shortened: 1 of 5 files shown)


Nalum:
@tomhuang12 @chanwit just thinking, would it make sense to have `imagePullSecrets` set for each `ServiceAccount` separately or are we okay with a global applied to both (as is in this PR)?

tomhuang12 (commented):
I think we can use the same secret for now since they are most likely stored in the same registry. Just a minor nit.

tomhuang12 (approved):
LGTM! Thank you!

PR #205, opened 2022-04-12
Swap from Label to Annotation

### LINKED ISSUE #204: Error "must be no more than 63 characters"
I'm getting the following error when I try to have my Terraform run:

```
error saving plan secret: rpc error: code = Unknown desc = error recording plan status: Secret "tfplan-default-testing" is invalid: metadata.labels: Invalid value: "plan-1a4dc9d874988bb51646864196d008f40816bb94c8ca54d3426c0c183f262702": must be no more than 63 characters
```

The Terraform script is very simple, just 1 output, for testing purposes:

```hcl
output "testing" {
  value = "Hello World"
}
```

Nalum:
Closes #204 

Also moved a couple of `consts` out from function scope and into package scope as their values were used in a few places.

### DIFF
--- a/controllers/tc000010_no_outputs_test.go
+++ b/controllers/tc000010_no_outputs_test.go
@@ -153,7 +153,7 @@ func Test_000010_no_outputs_test(t *testing.T) {
 			return nil
 		}
 		return map[string]interface{}{
-			"SavedPlan":             tfplanSecret.Labels["savedPlan"],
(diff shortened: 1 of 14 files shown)


Nalum:
@chanwit would it make sense to have a cross over here where we check for the label and if the label isn't there then check for the annotation?

tomhuang12:
Good catch on the issue and easy fix 🎉

chanwit:
all tests passed already. it's gone red because of the credential problem.

PR #223, opened 2022-05-06
Runner Pod EnvVars

Nalum:
Add support for environment variables on the Runner Pod set through the new `RunnerPodSpec` in the `RunnerPodTemplate`.

Also adds pass through of the envvars `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` from the controller to the runner. If these variables are set on the `Terraform` resource they will override the variables being picked up from the controller.


chanwit:
Thank you so much @Nalum for implementing this PR.
I rebased, squashed, and fixed the order of test cases in #232 to make this PR work.

Closed as reworked by #232

Nalum:
@chanwit awesome :+1: will review the diff and would be good to walk through it some time to get a better understanding of it

Thank you so much for the support !! :heart:

PR #232, opened 2022-05-21
Support environments variables for runner pod specification

chanwit:
Rework of #223 

Add support for environment variables on the Runner Pod set through the new `RunnerPodSpec` in the `RunnerPodTemplate`.

Also adds pass through of the envvars `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` from the controller to the runner. If these variables are set on the `Terraform` resource they will override the variables being picked up from the controller.

Rebased, squashed and rewrote tests.

Signed-off-by: Luke Mallon (Nalum) <luke.mallon@weave.works>
Signed-off-by: Chanwit Kaewkasi <chanwit@weave.works>
Co-authored-by: Luke Mallon (Nalum) <luke.mallon@weave.works>
Co-authored-by: Chanwit Kaewkasi <chanwit@weave.works>


stefanprodan (commented):
Several people at KubeCon were asking about HTTPS Proxy support for tf runners, so we need to make sure the terraform CLI does uses the env keys when they are specified in runner spec.

chanwit:
Thank you @stefanprodan 

Yes, thanks to the test case implemented by @Nalum, we already made sure the env vars were propagated to the Terraform binary.

stefanprodan (approved):
LGTM 

Thanks @chanwit for solving the proxy issues in record time 🥇

PR #236, opened 2022-05-22
revert back to go get as go install currently breaks e2e

chanwit:


chanwit:
Hi @o6uoq @Nalum 
please use Go 1.17.x only for the development as `go install` required by Go 1.18 breaks our E2E.

chanwit:
LGTM

PR #266, opened 2022-07-18
Fix IRSA annotation example in use_cases.md docs

benreynolds-drizly:
This PR fixes an example in the documentation which provides the incorrect annotation for applying IRSA to the `tf-controller` ServiceAccount.


chanwit:
Thank you so much, @benreynolds-drizly!

PR #267, opened 2022-07-18
Correct IRSA use case docs to annotate runner instead of controller

benreynolds-drizly:
Annotating the controller does not expose the AWS credentials to the AWS Terraform provider. The credentials need to be provided on the runner.


chanwit:
Yes, that's correct for the new runner architecture.
Thank you for the fix, @benreynolds-drizly!

chanwit (approved):

Issue #250, opened 2022-06-17
Terraform State Lock not released


Nalum:
We've hit an issue where the State lock was not released due to the runner pod not shutting down correctly (assumption here) due to a node refresh on an EKS Cluster. So we ended up with a Kubernetes `Lease` that was still held.

```
$ kubectl -n flux-system get lease
NAMESPACE     NAME                      HOLDER                                AGE
...
flux-system   lock-tfstate-default-sg1                                        3d7h
flux-system   lock-tfstate-default-sg2  f2ab685b-f84d-ac0b-a125-378a22877e8d  3d7h
```

The `Terraform` resource was reporting the following:

```
error running Plan: rpc error: code = Unknown desc = error acquiring the state lock: Lock Info: ID: f2ab685b-f84d-ac0b-a125-378a22877e8d Path: Operation: OperationTypePlan Who: controller@sg2-tf-runner Version: 1.1.9 Created: 2022-06-15 22:27:10.244794577 +0000 UTC
```

Which was repeated in the `tf-controller` and `tf-runner` logs. We were able to get things running again by deleting the `Lease` in question.

So the ask here is to support the [Terraform force-unlock](https://www.terraform.io/language/state/locking#force-unlock) feature.

I think the best approach might be to add a new field to the [`TerraformSpec`](https://weaveworks.github.io/tf-controller/References/terraform/#infra.contrib.fluxcd.io/v1alpha1.TerraformSpec), e.g. `forceUnlock`, the value of which should be the `Lock Info: ID: [value]`, in the above example this would be `forceUnlock: f2ab685b-f84d-ac0b-a125-378a22877e8d`. This ID corresponds to the key [`spec.holderIdentity`](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/lease-v1/#LeaseSpec) on a  `Lease`.

The `tf-controller`/`runner` should be tasked with acting on this, allowing for unlocking other Terraform backends that support locking if they become supported by the `tf-controller`.

This could then be supported in the `tfctl` e.g. `tfctl force-unlock f2ab685b-f84d-ac0b-a125-378a22877e8d` which would set the field on the Terraform resource and force a reconciliation. Is there a potential race condition here?

Nalum:
[...]

PR #268, opened 2022-07-19
Terraform Force Unlock

Nalum:
This is adding the ability to force unlock a terraform state, see #250 for more information


Nalum:
@chanwit just dropping a comment here so I don't forgot, if possible we need to get hashicorp/terraform-exec#268 merged or have your fork updated to at least v0.16.0 as that is when the force-unlock feature was added to the exec package.

chanwit (approved):

chanwit:
Thank you so much @Nalum !!

PR #282, opened 2022-08-08
support oci repository

### LINKED ISSUE #283: Support OCI Repository
Flux's Source controller v0.26.0+ introduces a new kind of source, called OCIRepository.
It would be great for us to support this source.

```yaml
---
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: OCIRepository
metadata:
  name: helloworld-oci
spec:
  interval: 1m
  url: oci://ghcr.io/tf-controller/helloworld
  ref:
    tag: main
---
kind: Terraform
metadata:
  name: helloworld-oci-auto-approve
spec:
  interval: 10s
  approvePlan: "auto"
  path: ./
  sourceRef:
    kind: OCIRepository
    name: helloworld-oci
```

chanwit:
This PR supports the OCIRepository source from the Source controller v0.26.0+

Fixes #283 


config/testdata/source/source.yaml
```diff
@@ -1,9 +1,20 @@
...
\ No newline at end of file
+    branch: main
+---
+apiVersion: source.toolkit.fluxcd.io/v1beta2
+kind: OCIRepository
+metadata:
+  name: helloworld-oci
+spec:
+  interval: 1m
+  url: oci://ghcr.io/tf-controller/helloworld
```
stefanprodan, line 18: I think you need to re push the artifact using the CLI built from the latest commit in the OCI branch:

```
flux list artifacts oci://ghcr.io/tf-controller/helloworld
ARTIFACT                             	DIGEST                                                                 	SOURCE	REVISION 
ghcr.io/tf-controller/helloworld:main	sha256:83eca7989ac1316ee764e10f6aa8690ba7b110cb1ea6dd5d1e2a1ce8c6d9e9ef	
```

The source and revision are missing.

chanwit: Thank you @stefanprodan.

I remember being told by you that the CLI was working with YAML only so I went for `crane` to package the TF module into the image. Does this imply that I can use Flux CLI to also package any kind of files?

stefanprodan: The Flux CLI can bundle anything, it doesn't look at files at all. Maybe in the future we'll add a dedicated flag for validating Kubernetes manifests, but this will be optional and not enabled by default.

chanwit: @stefanprodan I used Flux CLI to repack and also did a `cosign`.

```
flux push artifact oci://ghcr.io/tf-controller/helloworld:main \
  --path=./ --source=https://github.com/tf-controller/helloworld \
  --revision=HEAD

COSIGN_EXPERIMENTAL=1 cosign sign ghcr.io/tf-controller/helloworld:main
```

Looks great so far.

stefanprodan: Using HEAD as the revision is so bad lol

Here is how you should use it:
```
  flux push artifact oci://ghcr.io/tf-controller/helloworld:$(git rev-parse --short HEAD) \
	--path="./path/to/local/manifests" \
	--source="$(git config --get remote.origin.url)" \
	--revision="$(git branch --show-current)/$(git rev-parse HEAD)"

  flux tag artifact oci://ghcr.io/tf-controller/helloworld:$(git rev-parse --short HEAD) --tag=main
```
First you create an immutable artifact tagged with the Git SHA, then you tag it as main. Every time you make changes, you push a new artifact they you tag it as main.

chanwit: lol, that's my intention to add an arbitrary word as the revision.

stefanprodan: The OCI workflow mimics Git, you have immutable artifacts that represent Git commits, and mutable tags that represent Git branches. Flux follows a mutable tag instead of a branch, to know at which version you are in-cluster, you would lookup the OCI revision you pass at `flux push`. If people will set random things like HEAD in `--revision` it will be impossible to trace back changes.

chanwit: Gotcha. The snapshot representation and the traceability are really making sense. Good to know that the OC-based workflow is still related to Git. This is a very important feature.

stefanprodan (approved):
LGTM

Thanks @chanwit 🏅

PR #303, opened 2022-08-17
feat: support for targets (#270)

### LINKED ISSUE #270: Add support for terraform targets
I would like to see support for adding terraform targets.  I imagine something like

```
kind: Terraform
metadata:
  name: vpc
  namespace: flux-system
spec:
  ...
  path: ./terraform/test/us-west-2/dev/vpc
  targets:
  - module.dev_vpc
  varsFrom:
  - kind: Secret
    name: aws-credentials
  ...
```
When initially creating the infrastructure we would have the targets section commented out.  But afterwards we would want to restrict terraform to a set of targets.  This will allows us to do creates and deletes of VPCs and clusters without losing the EIPs in the original tf config.  While we can add lifecycle blocks to the EIPs, this prevents the entire tf config from being destroyed as terraform will error out about the EIPs and their prevent deletion lifecycles.

akselleirv:
Closes #270


chanwit:
Here's the error log in case you want to examine it.
https://github.com/weaveworks/tf-controller/runs/7884680235?check_suite_focus=true#step:6:1082

PR #308, opened 2022-08-22
feat: introduce helm-docs for consistency

oliverbaehler:
Signed-off-by: Oliver Bähler <oliverbaehler@hotmail.com>

Hi Guys, With this PR i added `helm-docs` for consistent value documentation. If you don't like the change, you can close the pr :).

Makefile:
  * Added `make helm-docs` to generate doc locally
  * Added `make helm-lint` to lint chart locally

Workflow:
  * Added diff to verify documentation is consistent with values

Helm Values:
  * Added `podLabels`
  * Added `priorityClassName`


chanwit:
Thank you so much @oliverbaehler 
I'll take a look at it shortly!

tomhuang12:
Love this! Thank you 🏅 . Just a minor nit to bump the chart version to pass the helm lint.

oliverbaehler:
@tomhuang12 thanks, rebase everything. @tomhuang12 i also added that the tag is used from `$.Chart.AppVersion` so you don't have to bump the tag versions in the values.yaml for both the controller and runner image everytime

charts/tf-controller/templates/deployment.yaml
```diff
@@ -46,12 +49,12 @@ spec:
             fieldRef:
               fieldPath: metadata.namespace
         - name: RUNNER_POD_IMAGE
-          value: "{{ .Values.runner.image.repository }}:{{ .Values.runner.image.tag}}"
+          value: "{{ .Values.runner.image.repository }}:{{ default $.Chart.AppVersion .Values.runner.image.tag }}"
```
tomhuang12, line 52: I don't think you need `$` as it is not in a particular scope and should be able to access `.Chart` without it.

charts/tf-controller/templates/deployment.yaml
```diff
@@ -46,12 +49,12 @@ spec:
...
         {{- range $key, $value := .Values.extraEnv }}
         - name: {{ $key }}
           value: {{ $value }}
         {{- end }}
-        image: "{{ .Values.image.repository }}:{{ .Values.image.tag}}"
+        image: "{{ .Values.image.repository }}:{{ default $.Chart.AppVersion $.Values.image.tag }}"
```
tomhuang12, line 57: Same here

tomhuang12 (commented):
Thank you! Didn't think about that one. Just minor nit.

chanwit:
We keep releasing, and the chart version causes an unnecessary error. 
I think I shall merge this PR, and we can continue from there.

chanwit (approved):
LGTM
thank you @oliverbaehler :1st_place_medal:

PR #310, opened 2022-08-23
Update ForceUnlock

Nalum:
Update the ForceUnlock feature to follow a similar pattern to the Plan feature.

We've added a new field to the Status:

```yaml
status:
  lock:
    lastApplied: [string]
    pending: [string]
```

ForceUnlock will only be called under the following circumstances:
- `spec.tfstate.forceUnlock == "yes" && spec.tfstate.lockIdentifier == status.lock.pending && status.lock.pending != ""`.
- `spec.tfstate.forceUnlock == "auto" && status.lock.pending != ""`.

This also sets up the Condition Types as constants in the `v1alpha1` package.

Previously I also setup two conditions related to the state lock, these are not combined into the condition type `ConditionTypeStateLocked` with two reasons `TFExecLockHeldReason` and `TFExecForceUnlockReason`.


chanwit (approved):
LGTM 
thank you so much @Nalum :1st_place_medal: !!

Issue #322, opened 2022-08-29
Supply Git credentials for runner


itamar-marom:
In the case of using a module that is stored in a git repository (self-hosted GitLab for example), the runner requires a username and password for cloning this module. Currently, it looks like there is no such way.

Log:
```shell
{"level":"error","ts":"2022-08-29T09:44:26.811Z","logger":"controller.terraform","msg":"error in terraform setup","reconciler group":"infra.contrib.fluxcd.io","reconciler kind":"Terraform","name":"NAME","namespace":"NS","error":"error running Init: rpc error: code = Unknown desc = exit status 1
Error: Failed to download module
Could not download module \"MODULE\" (main.tf:1) source code from
"git::https://HOSTED_GIT_ADDR/MODULE_PATH.git\":
error downloading
'https://HOSTED_GIT_ADDR/MODULEPATH.git':
/usr/bin/git exited with 128: Cloning into '.terraform/modules/MODULE'...
fatal: could not read Username for 'https://HOSTED_GIT_ADDR': No such
device or address
Error: Failed to download module
Could not download module \"MODULE\" (main.tf:1) source code from
\"git::https://HOSTED_GIT_ADDR/MODULE_PATH.git\":
error downloading
'https://HOSTED_GIT_ADDR/MODULE_PATH.git':
/usr/bin/git exited with 128: Cloning into '.terraform/modules/MODULE'...
fatal: could not read Username for 'https://HOSTED_GIT_ADDR': No such
device or address"}
```

I want to be able to pass it an access token, username, password

chanwit:
Thank for reporting this @itamar-marom 
how did you normally pass those credentials in?

itamar-marom:
With git credentials helper or running from GitLab CI in my case(which takes care of that), but there might be a better way.
https://www.terraform.io/language/modules/sources#generic-git-repository

I'll try to think on some ideas

bivas:
@chanwit I created a PR suggesting a way to run init commands to the runner container. The user can provide a command to run and by that @itamar-marom can run `git config` commands to setup his Git creds

chanwit:
Thank you @bivas 

Letting users run arbitrary commands is too risky for remote execution exploits.
A better way might be mapping Secrets data to file contents instead.

bivas:
I agree. Mapping is a much better idea.

chanwit:
That won't work because we have a volume mapping for the whole `/home/runner`.

[...]

PR #326, opened 2022-08-31
adding ability to run pre init commands in runner container

bivas:
Addresses https://github.com/weaveworks/tf-controller/issues/322


tomhuang12:
@chanwit this looks like a security concern to me. WDYT?

chanwit:
It is.

We did discuss about the concern with @bivas , the PR author.

@bivas shall we close this one please as you are already coming up with a new idea + implementation?

bivas:
@chanwit yes, we're preparing another PR that was discussed in the issue above

PR #335, opened 2022-09-05
Fix the test to actually test the selector changes

Nalum:
This PR:

- sets up the `Terraform` resources  in the `000310` tests using yaml strings rather than building up the object in Go.
- fixes the `000310` tests so that they are checking the generated `PodSpec` that the reconciler creates.

This break was called out in #331


chanwit (approved):
LGTM

PR #346, opened 2022-09-09
Add optional workspace field

k0da:
This commit adds support for selecting given Terraform workspace.


api/v1alpha1/terraform_types.go
```diff
@@ -85,6 +85,9 @@ type TerraformSpec struct {
 	// +optional
 	BackendConfigsFrom []BackendConfigsReference `json:"backendConfigsFrom,omitempty"`
 
+	// +optional
+	Workspace string `json:"workspace,omitempty"`
```
tomhuang12, line 89: I feel like `Workspace` should be part of `BackendConfig`

tomhuang12:
Thank you for your contribution 🥇! We will take a look shortly.

api/v1alpha1/terraform_types.go
k0da: Isn't BackendConfig directly mapped into backend {} terraform's config?

k0da: I see now, I agree it should be part of BackendConfigSpec

tomhuang12: That's true. I think it can be in `BackendConfig` since it is really a feature supported by a Terraform `backend`. However, it is also a top level command line option of `terraform`. I can go either way on this.

k0da: Moved to BackendConfig

tomhuang12:
Overall I think this looks great. Thanks so much for your contribution!

I am wondering how to make this feature more useful. Without this, user could technically mimic a workspace using different `Terraform` resources referencing the same repository but possibly same or different branches to achieve the [use cases](https://www.terraform.io/language/state/workspaces#when-to-use-multiple-workspaces) Terraform describes workspace is for. They will produce different state files which also mimics the [Workspace internals](https://www.terraform.io/language/state/workspaces#workspace-internals) described by Terraform. 

Doesn't mean that this addition isn't useful. Maybe users may already have workspaces set up and want to use them instead of creating new states. 

Maybe I am totally missing the point here but wanted to hear your use case on this @k0da.

k0da:
@tomhuang12 thanks for feedback. My use case is exactly moving existing machinery into tf-controller. Yes workspace is just a copy of a state. And technically it can be achieved by separate state object. When using s3 backend workspace reflected as subfolder under bucket path/env:/<workspacename> .

So I can't really implement my case without this patch.

tomhuang12:
Thanks for the explanation. The test is failing. Perhaps the grpc needs to be re-generated?

k0da:
weird they are generated:
```
runner/runner.pb.go
runner/runner.proto
runner/runner_grpc.pb.go
```

k0da:
Also test passing for me locally

controllers/terraform_controller.go
```diff
@@ -923,6 +923,14 @@ terraform {
...
 	log.Info("tfexec initialized terraform")
 
+	workspaceRequest := &runner.WorkspaceRequest{
+		TfInstance: tfInstance,
+		Terraform:  terraformBytes,
+	}
+	workspaceReply, _ := runnerClient.SelectWorkspace(ctx, workspaceRequest)
+
+	log.Info(fmt.Sprintf("workspace select reply: %s", workspaceReply.Message))
+
```
Nalum, line 933: I suggest you put a check in place either here or in the `SelectWorkspace` runner function.

runner/server.go
```diff
@@ -359,6 +359,38 @@ func (r *TerraformRunnerServer) Init(ctx context.Context, req *InitRequest) (*In
...
+	if ws != "" {
+		if err := r.tf.WorkspaceNew(ctx, ws, wsOpts...); err != nil {
+			log.Info("workspace new err, %s", err.Error())
+		}
+		if err := r.tf.WorkspaceSelect(ctx, ws); err != nil {
+			err := fmt.Errorf("failed to select workspace %s", ws)
+			log.Error(err, "workspace select error")
+			return nil, err
+		}
+	}
```
Nalum, line 390: Or here that checks you have a `terraform.Spec.BackendConfig` to work with, this is a pointer so will be `nil` where not explicitly set.

Nalum (changes requested):
Did you run all the tests or just `tc000020`?

k0da:
Thanks for review. Amended changes

Nalum:
Thanks for the changes, looks good and tests are passing. I'm wondering if we should be making the request if there is no workspace set. What do you think @chanwit @tomhuang12 ?

chanwit:
The default value of the workspace is "default".

Setting the different values to the `workspace` will affect many places. For example, we have `tfstate-default` and `tfplan-default` hardcoded everywhere, and they must be `tf-${worspace}` when a different workspace is specified.

k0da:
> The default value of the workspace is "default".
> 
> Setting the different values to the `spec.workspace` would affect many places.
> 
> For example, we have `tfstate-default` and `tfplan-default` hardcoded everywhere, and they must be `tf-${worspace}` when a different workspace is specified.

Thanks! I addressed this.

k0da:
@chanwit Hi, could you please take a look again? I hope I addressed all review points. Thanks!

chanwit:
Thank you @k0da 
could you please take a look at the logs,
here's a failing one: https://github.com/weaveworks/tf-controller/actions/runs/3035810853/jobs/4904144628#step:6:198

k0da:
@chanwit I moved Workspace out of BackendConfig as it breaks backend configuration. Now tests are passing.

k0da:
@chanwit I think Trivy scan should be handled separately

chanwit:
Thank you so much @k0da 
I'll try it on my local box and get back to you shortly.

chanwit (approved):

chanwit:
LGTM thank you @k0da :1st_place_medal:

chanwit:
It's in v0.13.0-rc.2.

Please install it via the rc.yaml.

PR #359, opened 2022-09-23
Fix crash if workspace select failed.

k0da:


chanwit (approved):
LGTM
Thank you @k0da :1st_place_medal:

PR #367, opened 2022-09-28
implement the values variable mapping

### LINKED ISSUE #366: Implement spec.values as a special variable
`spec.values` would give us a good level of abstraction (something like Crossplane, but better) without having a ton of CRDs.
Together with a newly developed package generator (tofu-jet), we would be allowed to express something like this:
```yaml
kind: Terraform
metadata:
  name: my-ec2-policy
  namespace: flux-system
spec:
  values:
    name: my-ec2-policy  
    description: "My EC2 policy"
    policy: |
      {
        "Version": "2012-10-17",
        "Statement": [
          {
                "Effect": "Allow",
                "Action": [
                    "ec2:*",
                ],
                "Resource": "*"
          }
        ]
      }
  path: ./aws_iam_policy
  sourceRef:
    kind: OCIRepository
    name: aws-primitive-modules-v4.32.0
```

Flux OCIRepository would become our packaging system to deliver this new kind of modules.
We would also be able to support AWS primitive modules out of the box.

chanwit:
Fixed #366 

This PR is straightforward. It introduces a new field `.spec.values`. This field is responsible for mapping the YAML of `spec.values` to the Terraform variable `values` which must be defined in a Terraform, and have the same structure. The `values` variable must be an object.

The same approach has been implemented for a long time in Flux's `HelmRelease`.

Implementing this allows us to create a new kind of abstraction, similar to what is currently offered by Crossplane, but without adding any special kind of CRDs.

This feature will be used together with OCI-based Terraform modules generated by a new tool called Tofu-Jet.
We have recently released an alpha version of AWS primitive modules, which is supported by this feature, [here](https://github.com/tf-controller/aws-primitive-modules)


chanwit:
@Nalum requested you to review so that I could explain the new feature here.

Nalum (dismissed):
Thanks @chanwit I understand what is going on now, this seems really nice especially with the tofu-jet work!

Issue #389, opened 2022-10-05
Partial backend configuration clashes with `backendConfig.customConfiguration`


adamstrawson:
Our Terraform modules use a partial configuration, which is recommended within the Terraform [documentation](https://www.terraform.io/language/settings/backends/configuration#partial-configuration) 

For example;

```
terraform {
  backend "gcs" {}
}
```

Locally, we use Terragrunt to populate the backend, and within cluster we use [/isaaguilar/terraform-operator](https://github.com/isaaguilar/terraform-operator) (which is what we're currently using, and hoping to migrate to `tf-controller`) `terraform-operator` supports the ability to pass a custom backend which replaces the partial configuration above. 

However, when setting the below example, tf-runner fails due to a duplicate backend. 

```
---
apiVersion: infra.contrib.fluxcd.io/v1alpha1
kind: Terraform
metadata:
  name: foo
  namespace: bar
spec:
  interval: 5m
  approvePlan: auto
  sourceRef:
    kind: GitRepository
    name: helloworld
    namespace: bar
  backendConfig:
    customConfiguration: |
        backend "gcs" {
          bucket = "terraform-state"
          prefix = "my-path"
          impersonate_service_account = "a-service-account
        }
```

tf-runner log for above resource:
```
There are some problems with the configuration, described below.

The Terraform configuration must be valid before initialization so that
Terraform can determine which modules and providers need to be installed.

Error: Duplicate backend configuration
 on main.tf line 2, in terraform:
 2: backend "gcs" {}

A module may have only one backend configuration. The backend was previously
configured at generated_backend_config.tf:3,3-16.
```

Does tf-controller not support partial configuration on backends, or is there an error in the documentation around how this should be formatted?

chanwit:
Hi @adamstrawson 

Thank you for reporting the issue. We don't have a mechanism to support your use case yet.
But it's really a feature we would support. Maybe we would generate an override.tf instead?

adamstrawson:
That would work @chanwit - I've submitted a PR (#392) with a solution for this issue.

aristapimenta:
hi @adamstrawson, thanks for spotting and fixing this issue!

I'm currently working on a PoC and am trying to use tf-controller with GCS backend as well. could you pls provide more details or pointers on how do you do that via `impersonate_service_account`?

I'm currenty able to do something like this for authorizing the google terraform provider to manage GCP resources:

inside `main.tf`:
```hcl
variable "GOOGLE_CREDS" {
[...]

PR #392, opened 2022-10-07
Replace backendConfigPath with Override Files format

adamstrawson:
Terraform has special handling of any configuration file whose name ends in `_override.tf` or `_override.tf.json`. This special handling also applies to a file named literally `override.tf` or `override.tf.json`.

By renaming `backendConfigPath` to `backend_override.tf` it'll allow `backendConfig.customConfiguration` to be set and overwritten as expected when a backend is already defined in the Terraform Module.

Fixes: #389


tomhuang12 (approved):
It looks fine to me. To your specific use case, I would be careful with the [merging behavior](https://www.terraform.io/language/files/override#merging-behavior).

chanwit:
Thank you @adamstrawson :1st_place_medal:

Issue #297, opened 2022-08-16
Can't use different namespace


vbezhenar:
I installed tf-controller using the following flux files:

```
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: tf-controller
  namespace: tf-controller
spec:
  chart:
    spec:
      chart: tf-controller
      version: 0.4.3
      sourceRef:
        kind: HelmRepository
        name: tf-controller
  interval: 24h
```

I can see that it created service account `tf-runner` in namespace `tf-controller`:

```
kubectl get -A serviceaccounts |grep tf-runner
tf-controller            tf-runner                            1         12m
```

Now I'm trying to reconcile Terraform resource which is located in a different namespace.

```
apiVersion: infra.contrib.fluxcd.io/v1alpha1
kind: Terraform
metadata:
  name: users
  namespace: keycloak-system
spec:
  approvePlan: auto
  vars:
    - name: keycloak_administrator_password
      valueFrom:
        secretKeyRef:
          key: password
          name: keycloak-administrator
    - name: smtp_auth_password
      valueFrom:
        secretKeyRef:
          key: password
          name: smtp-auth
  interval: 24h
  sourceRef:
    kind: GitRepository
    name: cmctech-cluster-users
    namespace: flux-system
```

This does not work. I'm getting the following errors in controller:

```
{"level":"error","ts":"2022-08-16T12:19:57.013Z","logger":"controller.terraform","msg":"unable to lookup or create runner","reconciler group":"infra.contrib.fluxcd.io","reconciler kind":"Terraform","name":"users","namespace":"keycloak-system","error":"pods \"users-tf-runner\" is forbidden: error looking up service account keycloak-system/tf-runner: serviceaccount \"tf-runner\" not found"}
{"level":"error","ts":"2022-08-16T12:19:57.013Z","logger":"controller.terraform","msg":"Reconciler error","reconciler group":"infra.contrib.fluxcd.io","reconciler kind":"Terraform","name":"users","namespace":"keycloak-system","error":"pods \"users-tf-runner\" is forbidden: error looking up service account keycloak-system/tf-runner: serviceaccount \"tf-runner\" not found"}
```

I expect the controller to work across all namespaces in the cluster. Is it current limitation or did I miss some way to configure it? Should I install separate copy of tf-controller into every namespace with terraform resource?

artem-nefedov:
[...]

PR #430, opened 2022-10-17
Ability to create ServiceAccount for each runner namespace

adamstrawson:
Following on from https://github.com/weaveworks/tf-controller/issues/297

Rather than creating additional resources outside of the chart, this PR adds the ability to pass a list of namespaces that the runner may run within, and creates the appropriate ServiceAccount and ClusterRoleBinding resources within each namespace.

```
runner:
  serviceAccount:
    allowedNamespaces:
      - test
      - stable
      - stage
```


charts/tf-controller/templates/rbac.yaml
```diff
@@ -185,6 +185,8 @@ subjects:
 - kind: ServiceAccount
   name: {{ include "tf-controller.serviceAccountName" . }}
   namespace: {{ .Release.Namespace }}
+{{- $allowedNamespaces := append .Values.runner.serviceAccount.allowedNamespaces .Release.Namespace -}}
+{{- range $allowedNamespaces }}
```
Nalum, line 189: Does this not create a conflict of duplicate objects or overwrite?

I think it would be better to move this `range` down under the `subjects` line.

Nalum (approved):
LGTM

Nalum:
Looks like the docs need to be updated  https://github.com/weaveworks/tf-controller/actions/runs/3273201286/jobs/5385113192

adamstrawson:
> Looks like the docs need to be updated https://github.com/weaveworks/tf-controller/actions/runs/3273201286/jobs/5385113192

Docs pushed - sorry, we let GHA generate our docs, too used to that process!

Nalum (approved):
LGTM

PR #433, opened 2022-10-24
 Fix runner service account template not returning multiple documents when defining allowedNamespaces

skeletorXVI:
Fixes the following issue:

```
Helm install failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors:\n  line 16: mapping key \"apiVersion\" already defined at line 2\n  line 17: mapping key \"kind\" already defined at line 3\n  line 18: mapping key \"metadata\" already defined at line 4
```

### DIFF
--- a/charts/tf-controller/Chart.yaml
+++ b/charts/tf-controller/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: tf-controller
 description: The Helm chart for Weave GitOps Terraform Controller
 type: application
-version: 0.8.8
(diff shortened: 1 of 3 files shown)


chanwit:
Thank you @skeletorXVI so much for your contribution.

@tomhuang12 could you please help review it?

skeletorXVI:
The failing pipeline seems unreleated to my pull request

tomhuang12 (approved):

tomhuang12:
Good catch!

Issue #436, opened 2022-10-26
AppliedSuccefully doesn't set condition properly


k0da:
``` 
 - lastTransitionTime: "2022-10-26T10:35:55Z"
    message: 'Applied successfully: awesome-repo/f368f993dc74ca0f5a6588e24b356f518c8b7ce8'
    reason: TerraformAppliedSucceed
    status: Unknown
    type: Ready
```
status should be True

k0da:
https://github.com/weaveworks/tf-controller/blob/main/api/v1alpha1/terraform_types.go#L550
explicitly sets unknown

chanwit:
Hi @k0da 
Thank you for that.

We're working on fixing this and other related issues in a different PR.

chanwit:
Fixed by #438 and #440 (v0.13.0)
Please feel free to re-open it if you still see the error.

PR #437, opened 2022-10-26
Fix terraform applied condition status

k0da:
Fixes https://github.com/weaveworks/tf-controller/issues/436


chanwit:
@k0da
It was intended to be Unknown so that we could summarize the real action later, here:
https://github.com/weaveworks/tf-controller/blob/70aa227c4414e5913bfb0523e57a0c3ca9c8f822/controllers/tf_controller_reconcile.go#L183-L191

chanwit:
Please test it in the recent v0.13.0 release.

PR #446, opened 2022-10-28
Add runner-aws image

k0da:

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -115,6 +115,25 @@ jobs:
             org.opencontainers.image.revision=${{ github.sha }}
             org.opencontainers.image.version=${{ steps.prep.outputs.VERSION }}
             org.opencontainers.image.created=${{ steps.prep.outputs.BUILD_DATE }}
+      - name: Publish multi-arch tf-runner-aws container image
+        uses: docker/build-push-action@v2
+        with:
+          push: true
+          no-cache: true
+          builder: ${{ steps.buildx.outputs.name }}
+          context: .
+          file: ./runner-aws.Dockerfile
+          platforms: linux/amd64 #,linux/arm/v7,linux/arm64
+          tags: |
+            ghcr.io/weaveworks/tf-runner-aws:${{ steps.prep.outputs.VERSION }}
+            ghcr.io/weaveworks/tf-runner-aws:latest
+          labels: |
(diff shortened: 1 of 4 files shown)


chanwit:
Thank you for this PR, @k0da 

Not like Azure, 

In case of Azure, we needed the CLI for some resource provisions, but I'm not very sure why do we need a specific image for AWS?

cc @tomhuang12 @Nalum if you have some more info about AWS.

tomhuang12:
Change it is right. I am wondering why it needs aws cli.

k0da:
Kubectl provider is using aws cli to get access tokens from EKS

k0da:
https://registry.terraform.io/providers/gavinbunney/kubectl/latest/docs#exec-plugin-support

chanwit:
I see. We don't have enough bandwidth to maintain many vendor-specific images.
Maybe we would create the separate image-contrib repo to host the growing number of images. wdyt?

k0da:
@chanwit what we can potentially do is to base custom images on top of runner and excluder TF specifics build from it. So it only will have azure/aws cli install steps

k0da:
I went ahead and implemented eks get token functionality within small golang program, this reduces install steps into a single curl https://github.com/AbsaOSS/aws-get-token .

Any ideas how we can proceed making provider specific images as part of official pipelines?

yitsushi:
> Any ideas how we can proceed making provider specific images as part of official pipelines?

Sorry for the late answer. I don't think we want to go to that direction and manage any kind of "official" or "officially approved contrib" repositories for various custom images. Most projects stopped doing that because an "aws runner" is good for the one who created first, but the next one wants something else built into the runner image and that repo would end up with 10+ aws runners, 10+ gcloud runners, 10+ something-else runners, etc.

Like a other projects, people can create repos, they can use them, just like github actions (or DroneIO actions as containers), everyone can find what they want or create a new one if nothing fits their needs.

Starting points:
- How to use custom images: https://weaveworks.github.io/tf-controller/use_tf_controller/to_provision_resources_with_customized_Runner_Pods/
- Change runner pod image globally: https://github.com/weaveworks/tf-controller/blob/main/charts/tf-controller/values.yaml#L119
- An example of a custom image: https://github.com/weaveworks/tf-controller/blob/main/runner-azure.Dockerfile

Issue #336, opened 2022-09-06
Different solution for storing Terraform plans


itamar-marom:
Storing the Terraform plan in a secret has two limitations:
- Large plans cannot be stored in secret due to the secret limit
- Storing those secrets (on a large scale) in the etcd can cause some performance issues

We should find a different store for that.

chanwit:
Thank you @itamar-marom 

We would use a similar configuration as `backend` for storing plans.

tomhuang12:
We recently released a version that can configure a [custom backend](https://weaveworks.github.io/tf-controller/use_cases/#use-custom-backend). Please let us know if this meets your needs.

bivas:
@tomhuang12 this issue is for plan storage (binary and JSON format) not Terraform state (which covered by custom backend as you mentioned).
Current plan storage is Kubernetes Secret and it's problematic with large plans (we have plans averaging at 6.5MB as JSON format) and a different storage is required (e.g. PVC)

itamar-marom:
Hi @chanwit @tomhuang12

## Our goal:  
Be able to store Terraform plan files in persistent storage.  
- Storage type will be agnostic
- Terraform Controller can be run locally

## Our suggested solution:
Use Kubernetes’ `PersistentVolume` and `PersistentVolumeClaim` storage abstraction solutions.  We want to release this solution in two steps.

For production, any supported `StorageClass` can be used.   
A PVC with the `EmptyDir` option can be attached for a local run.

In each of those solutions, a PVC will be attached to the runner pod.  

### Version 1
`Terraform` will have a plan configuration field in its spec with a `claimName` field - a reference to a PVC to use. Terraform controller will then attach the PVC to the relevant runner pod. 

Sample API:
```yaml
# We create the PVC ourself
apiVersion: infra.contrib.fluxcd.io/v1alpha1
kind: Terraform
metadata:
    name: 
spec:
    planConfig: # Configuration related to Terraform plan
        storage:
	    claimName: efs-claim # PVC name to use in runner
    sourceRef: # GitRepository as source
  	kind: GitRepository
        name: helloworld
        namespace: flux-system
```

### Version 2
Instead of user has to create the PVC on it's own, we would the Terraform controller to have a similar solution to Kubernetes `StatefulSet` with its `volumeClaimTemplate` spec. In this stage, the Terraform controller will create the PVC on its own for each runner.

Sample API: (We also talked about the possibility of making it more simple but less similar to StatefulSet's spec)
```yaml
[...]

PR #480, opened 2022-12-14
Feature/plan pvc support

itamar-marom:
#336 

With this PR, we can store terraform plans in Kubernetes `PersistentVolumeClaim`.  
The given PVC should exist already and is not part of the controller's responsibility.  
- Also supports all current options of `storeReadablePlan`

The runner will mount into a subPath that points to its own directory. The path looks like this:  
`/mnt/plan/terraform_controller/TERRAFORM_OBJECT_NAMESPACE/TERRAFORM_OBJECT_NAME/`  
This gives the security of a runner to mount only into its own object and also takes care the backend storage is organized and can serve other stuff then the terraform controller.

Example:  
```yaml
---
apiVersion: v1
kind: PersistentVolume
metadata:
  name: plan-config-claim-pv
  labels:
    type: local
spec:
  storageClassName: standard
  capacity:
    storage: 1Gi
  accessModes:
    - ReadWriteOnce
  hostPath:
    path: "/tmp/plan-test"
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: plan-config-claim-pvc
spec:
  storageClassName: standard
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
---
kind: Terraform
metadata:
  name: helloworld-plan-config-claim
spec:
  interval: 10s
  approvePlan: "auto"
  storeReadablePlan: "human"
  path: ./
  planConfig:
    storage:
      claimName: "plan-config-claim-pvc"
  sourceRef:
    kind: GitRepository
    name: helloworld
```

### DIFF
--- a/api/v1alpha1/terraform_types.go
+++ b/api/v1alpha1/terraform_types.go
@@ -797,7 +797,7 @@ func (in Terraform) GetClaimName() string {
 }
 
 func (in Terraform) GetPlanStorageMountSubDir() string {
-	return strings.Join([]string{PlanStorageMountSubDirBase, in.Name, in.Namespace}, "/")
+	return strings.Join([]string{PlanStorageMountSubDirBase, in.Namespace, in.Name}, "/")
 }
 
 // GetStatusConditions returns a pointer to the Status.Conditions slice.


itamar-marom:
Hi @yitsushi 
Are you into reviewing this also? If so I will fix the conflicts.
Link to issues: #336 #536

yitsushi:
You can, but I can't promise I'll have time to review in the near future as it requires changes we have to consider if we even want/can do.
1. It'll conflict with `main` when #1185 is merged, so you would have to resolve again.
2. It adds an extra resource to manage and we are already working on a solution to reduce that dependency or at least be able to reconcile is those resources are not available.

One of the main goals of the Workspace BLOB cache initiative is to prevent stuck resources when Source gets deleted before the Terraform resource. With this addition we would have to think about that extra dependency too, cache somewhere, manage what happens if it can't mount the resource yet still don't get stuck on a reconcile loop unable to even delete the resource without patching the `finalizer` and hacking the system hoping everything works out.

Any ideas, suggestions, opinions @chanwit ?

tyriis:
:wave: hi there, I am currently running in the issue, that my plan is to big. Just out of curiosity, what is wrong with the PR apart from the merge conflict?

manicole:
Hi, same here, any news on that ? Any workaround ? Thanks a lot !

itamar-marom:
HI :) 
Unfortunately, In the company I worked before (were I used the tofu controller) We decided to run with a Fork of ours that had some more features and improvements...

I am no longer active on this one

PR #487, opened 2023-01-04
Selector labels and custom labels had different indentation

geNAZt:
This lead to invalid YAML because when set the resulting labels would format to something like this:

```yaml
labels:
  customlabel: test
    selectorlabel: test
```


chanwit:
Thank you @geNAZt 
could you please:
- bump Helm chart version
- run helm doc generation Make target
then commit and push please.

tomhuang12:
Tested and LGTM! Thank you!

tomhuang12 (approved):

PR #494, opened 2023-01-10
feat: enable parallelism option for TF apply stage

siiimooon:
Hi,
 
 The parallelism option would come in handy when using providers that do not handle parallell operations gracefully.


chanwit:
Thank you for the PR @siiimooon 
It is a good addition. I'm taking a look at it and it would be possible to get merged soon.

runner/server.go
```diff
@@ -1005,6 +1005,10 @@ func (r *TerraformRunnerServer) Apply(ctx context.Context, req *ApplyRequest) (*
 		applyOpt = append(applyOpt, tfexec.Target(target))
 	}
 
+	if req.Parallelism != 0 {
```
chanwit, line 1008:
```suggestion
	if req.Parallelism > 0 {
```

chanwit: It's to prevent the user to specify a negative value.

chanwit (changes requested):

api/v1alpha1/terraform_types.go
```diff
@@ -212,6 +212,11 @@ type TerraformSpec struct {
 	// +optional
 	Targets []string `json:"targets,omitempty"`
 
+	// Parallelism limits the number of concurrent operations of Terraform apply step.
```
chanwit, line 215:
```suggestion
	// Parallelism limits the number of concurrent operations of Terraform apply step. Zero (0) means using the default value.
```

chanwit (changes requested):

siiimooon:
Thanks for the feedback, @chanwit. I have implemented both of your suggestions.

chanwit (approved):
LGTM!
Thank you so much @siiimooon :1st_place_medal:

PR #497, opened 2023-01-17
Fix error in docs examples

kingdonb:
You can see the effects of the unclosed codeblock here:

https://weaveworks.github.io/tf-controller/use_tf_controller/with_GitOps_dependency_management/


chanwit (approved):
Thank you @kingdonb 
LGTM!

Issue #347, opened 2022-09-09
error pending plan and plan's name in the secret are not matched


github-vincent-miszczak:
Hello,

I have 3 `terraform` resources managed by tf-controller.
2 of them are fine, one reports 
```
rpc error: code = Unknown desc = error pending plan and plan's name in the secret are not matched:  != plan-master-4246be8f0f8c0ffa86ea09b88d0000123a5562f0
```
When updating Git code, the updates do happen, but the message comes back with new commit ID.

Having a quick look, https://github.com/weaveworks/tf-controller/blob/3ae12dd2f2f82a5f89e3280ebb1e0a7b39393ecc/runner/server.go#L774 is an empty string, which looks fine has there no change to apply once applied successfully.

I need some help to fix this.

tomhuang12:
Could provide the secret metadata of the corresponding terraform plan? It should be start with `tfplan-default-`.

github-vincent-miszczak:
```
metadata:
  annotations:
    encoding: gzip
    savedPlan: plan-master-4246be8f0f8c0ffa86ea09b88d0000123a5562f0
  creationTimestamp: "2022-09-09T16:00:58Z"
  name: tfplan-default-xxx-apne1-xxx
  namespace: production
  ownerReferences:
  - apiVersion: infra.contrib.fluxcd.io/v1alpha1
    kind: Terraform
    name: xxx-apne1-xxx
    uid: c4819818-4300-4041-9142-99db53098c2b
  resourceVersion: "32325387"
  uid: 08d80990-7c48-4d11-9f49-3b63cb6d5692
```

This is similar to other working resources, example of a working one:
```
metadata:
  annotations:
    encoding: gzip
    savedPlan: plan-master-4246be8f0f8c0ffa86ea09b88d0000123a5562f0
  creationTimestamp: "2022-09-09T16:00:05Z"
  name: tfplan-default-xxxx-use1-xxx
  namespace: production
  ownerReferences:
  - apiVersion: infra.contrib.fluxcd.io/v1alpha1
    kind: Terraform
    name: xxx-use1-xxx
    uid: ccdcfb78-045e-49cc-93ce-0ec6cc231999
  resourceVersion: "32325065"
  uid: 451a240b-b501-49c3-a53e-7740cd872ec8
```

github-amine-kherbouche:
Hello,
any update on this issue?

enekofb:
i am facing a similar issue

```
[...]

PR #498, opened 2023-01-17
Force replan if cannot load tfplan correctly

tomhuang12:
Addresses #347


chanwit (approved):
LGTM

Thank you @tomhuang12 :100:

PR #514, opened 2023-02-06
add support for building multi-arch docker images

rparmer:
Update Dockerfiles and pipelines to build multi-arch variants of images

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -66,7 +66,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: ./Dockerfile
-          platforms: linux/amd64 #,linux/arm/v7,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
           tags: |
             ghcr.io/weaveworks/${{ env.CONTROLLER }}:${{ steps.prep.outputs.VERSION }}
             ghcr.io/weaveworks/${{ env.CONTROLLER }}:latest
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -66,7 +66,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: ./Dockerfile
-          platforms: linux/amd64 #,linux/arm/v7,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
```
chanwit, line 69: Please remove arm/v7 for the moment.
Production clusters would use only amd64 and arm64.

```suggestion
          platforms: linux/amd64,linux/arm64 # linux/arm/v7
```

.github/workflows/release.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
           builder: ${{ steps.buildx.outputs.name }}
           context: .
           file: ./runner.Dockerfile
-          platforms: linux/amd64 #,linux/arm/v7,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/arm/v7
```
chanwit, line 88:
```suggestion
          platforms: linux/amd64,linux/arm64 # linux/arm/v7
```

chanwit (changes requested):

.github/workflows/release.yaml
rparmer: done

chanwit (approved):
Thank you so much Russ!

PR #523, opened 2023-02-10
Allow customizing controller log encoding

### LINKED ISSUE #522: Allow tf-controller Log Encoding to Be Configurable
I have `v0.13.1` deployed for tf-controller with the `0.10.1` chart version. My organization uses `aws-for-fluent-bit` (since we use EKS) to ship event logs from clusters to Splunk Enterprise.

The logging format for tf-controller is hardcoded in the `deployment.yaml` template file, but it's commonly supported in the other gotk controller stack to be able to configure the `--log-encoding` arg to `console` aside from the default `json` value.

In my specific case, to get tf-controller's event logs "prettified" I would have to do some transformation on the Splunk Enterprise side of things. Without this prettification preprocessing before indexing in Splunk, you end up with JSON shoe horned into a String data type (in Splunk), which makes interpreting those events painful. It would be ideal to have the ability to specify the `--log-encoding` value as  `console` via a values.yaml override and substitution.

tomhuang12:
Add the ability to customize log encoding on the controller in the chart

Fixes #522

### DIFF
--- a/charts/tf-controller/Chart.yaml
+++ b/charts/tf-controller/Chart.yaml
@@ -2,5 +2,5 @@ apiVersion: v2
 name: tf-controller
 description: The Helm chart for Weave GitOps Terraform Controller
 type: application
-version: 0.10.1
+version: 0.10.2
 appVersion: "v0.13.1"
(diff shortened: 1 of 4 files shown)


chanwit (approved):
LGTM

PR #596, opened 2023-05-12
Add labels and annotations to outputted secret

### LINKED ISSUE #593: Add labels and annotations to outputted secrets
Is there a manner in which you can add labels and annotations to the outputted secret in a declarative manner? In a similar way that we can do with the 'podrunnertemplate'.

scott-david-walker:
Fixes #593 
Allows labels to be added to the created output secret with api defined in #593


chanwit (approved):
LGTM! Thank you @scott-david-walker :1st_place_medal:

PR #607, opened 2023-05-17
RFC-0001 Polling mechanism for the branch-based planner

### LINKED ISSUE #595: Write an RFC describing the security considerations for branch-based planner
Relates to #527 

The RFC should:
- Evaluate alternative approaches for communicating with git providers (i.e webhook vs polling) and their security posture
- Propose an approach and justify it
- Propose an architecture that supports the chosen approach

chanwit:
This RFC proposes a polling mechanism as a secure and portable solution for tracking changes in the
Branch-based Planner with GitHub. The proposed system reduces security risks associated with webhooks, which require
exposing a publicly accessible endpoint by having the Kubernetes cluster periodically query GitHub for changes.

Fixes #595 

Per the RFC process, the proposal can only be merged after it has been approved by either
-  two project maintainers, or
-  one project maintainer and one individual from the following: a Weaveworks Tech Lead, a Weaveworks Product Manager, or a Weaveworks Principal Engineer.

### DIFF
--- a/rfcs/0001-polling-mechanism/README.md
+++ b/rfcs/0001-polling-mechanism/README.md
@@ -1,0 +1,259 @@
+# RFC-0001 Polling mechanism for the branch-based planner
+
+**Status:** provisional
+
+<!--
+Status represents the current state of the RFC.
+Must be one of `provisional`, `implementable`, `implemented`, `deferred`, `rejected`, `withdrawn`, or `replaced`.
+-->
+
+**Creation date:** 2023-05-17
+
+**Last update:** 2023-05-17
+
+## Summary
+
+This RFC proposes a polling mechanism as a secure and portable solution for tracking changes in the
+Branch-based Planner with GitHub. The proposed system reduces security risks associated with webhooks, which require
+exposing a publicly accessible endpoint, by having the Kubernetes cluster periodically query GitHub for changes. The
+polling method also offers portability across other Git providers like GitLab and Bitbucket, as it avoids the complexity
+of dealing with varying webhook systems. The RFC also takes into account the potential drawbacks of the polling
+mechanism, such as latency and load on the GitHub server, and suggests careful configuration of the polling frequency.
+It further discusses considerations related to GitHub's API rate limit and recommends storing the GitHub token in a
+Kubernetes Secret for secure and efficient rate limit management.
+
+## Motivation
+
+In order to effectively track changes in a branch-based planner, it is essential to establish a solid and secure method
+of interaction between GitHub and the Kubernetes clusters. Two possible approaches are Webhooks and a polling mechanism.
+
+Webhooks are a common choice because of their real-time nature. However, they do introduce a considerable security risk.
+To use webhooks, a publicly accessible endpoint is required, which exposes the Kubernetes cluster to the outside world.
+While security measures can be put in place to protect this endpoint (such as using secure tunneling or authentication
+mechanisms), the exposure itself presents a risk. An attacker could potentially exploit vulnerabilities in the webhook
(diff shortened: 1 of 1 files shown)


yitsushi:
For polling, do we need an extra CRD that describes a repo to watch? Something like this:

```yaml
---
apiVersion: infra.contrib.fluxcd.io/v1alpha2
kind: TerraformBranchBasedPlanner
metadata:
  name: helloworld-tf
  namespace: flux-system
spec:
  interval: 1m
  sourceRef:
    kind: GitRepository
    name: helloworld
    namespace: flux-system
```

rfcs/0001-polling-mechanism/README.md
```diff
@@ -0,0 +1,259 @@
...
+On the contrary, the polling mechanism works largely the same way for any Git provider. We need to make requests to the
+provider's API to fetch the required data, which is typically available through similar RESTful APIs across different
+providers. Therefore, we can easily adapt our polling code to a new provider by changing the API endpoints and adjusting
+for any differences in the API responses.
+
+In addition, the polling mechanism allows us to control the rate of requests, which can be beneficial for dealing with
+rate limits imposed by different providers. With webhooks, the rate of incoming messages is determined by the activity
+on the Git provider, which could potentially overwhelm our system or cause us to exceed rate limits.
+
+## Proposed solution
```
yitsushi, line 82: For polling, do we need an extra CRD that describes a repo to watch? Something like this:

```yaml
---
apiVersion: infra.contrib.fluxcd.io/v1alpha2
kind: TerraformBranchBasedPlanner
metadata:
  name: helloworld-tf
  namespace: flux-system
spec:
  interval: 1m
  sourceRef:
    kind: GitRepository
    name: helloworld
    namespace: flux-system
```

chanwit: I would go with the simplest implementation first without introducing a new CRD.

yitsushi: Hmm... and what would be the "simplest implementation"?

chanwit: A simple polling, PR query, etc without CRD - that's what I meant.

yitsushi: Yes, but how the system will know what repositories it has to query for PRs?

chanwit: Would you please create a spike ticket and move this discussion there? It's a bit out of scope for this RFC.

yitsushi: Yes, but I don't think it's fully out of scope as it's probably a "drawback" as we have to either create a CRD or iterate over all Terraform resources on the cluster, parse their repo, aggregate the repositories (so no duplicates) and then we can query PRs. Compared to webhook where we know the repository from received payload.

chanwit: > Compared to webhook where we know the repository from received payload.

I mistook your concern then. 

We have that information in the code provided. See example 1.

yitsushi: In the provided code we have a hard-coded repository.
```go
prs, _, _ := client.PullRequests.List(ctx, "weaveworks", "tf-controller", nil)
```

yitsushi: ConfigMap sounds good.

chanwit: Scoping down to a set of specified repositories would also improve security.

Thank you for bringing this up @yitsushi :+1:

yitsushi (approved):
lgtm

yiannistri (approved):

PR #609, opened 2023-05-17
Setting up instance label for Runner pods

### LINKED ISSUE #547: Error: must be no more than 63 characters - limit on labels.
Hi! we are using tf-controller to provision kafka resources such as topics and schemas in Confluent Cloud, and it works wonderful. 

However, we have encountered an issue where the runner-pod is extracting the name of the pod to add the label. There is a 255 character limit on the name of the pod but only a 63 character limit on the label.

```
error":"Pod \"rolebinding-s99999-test-s99999-private-insight-usage-app-fct-tf-runner\" is invalid: metadata.labels
Invalid value: \"rolebinding-s99999-test-s99999-private-insight-usage-app-fct-tf-runner\": must be no more than 63 characters"}
```

In order for us to enforce uniqueness in topic-names and terraform CR's we need to have quite long names. 

Is there any possibility to change from using label to annotations for this specific label: https://github.com/weaveworks/tf-controller/blob/main/controllers/tf_controller_runner.go#L50

luizbafilho:
closes #547 

- Setting the `app.kubernetes.io/instance` label to `tf-runner-<short-rev>` to avoid hitting label size limits
- Fixing some tests assertions

### DIFF
--- a/controllers/tf_controller_runner.go
+++ b/controllers/tf_controller_runner.go
@@ -38,9 +38,14 @@ func getRunnerPodImage(image string) string {
 	return runnerPodImage
 }
 
-func runnerPodTemplate(terraform infrav1.Terraform, secretName string, revision string) v1.Pod {
+func runnerPodTemplate(terraform infrav1.Terraform, secretName string, revision string) (v1.Pod, error) {
 	podNamespace := terraform.Namespace
 	podName := fmt.Sprintf("%s-tf-runner", terraform.Name)
+	podInstance, err := runnerPodInstance(revision)
+	if err != nil {
+		return v1.Pod{}, err
+	}
+
 	runnerPodTemplate := v1.Pod{
 		ObjectMeta: v12.ObjectMeta{
 			Namespace: podNamespace,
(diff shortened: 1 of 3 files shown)


controllers/tf_controller_runner.go
```diff
@@ -462,3 +463,7 @@ func (r *TerraformReconciler) reconcileRunnerSecret(ctx context.Context, terrafo
 
 	return result.Secret, nil
 }
+
+func runnerPodInstance(revision string) string {
+	return fmt.Sprintf("tf-runner-%s", strings.Split(revision, ":")[1][0:8])
+}
```
chanwit, line 469: Please add some checks to make sure that we're not breaking the controller when a weird value being passed thru this function.
```suggestion
func runnerPodInstance(revision string) (string, error) {
        // please add some checks here
	return fmt.Sprintf("tf-runner-%s", strings.Split(revision, ":")[1][0:8])
}
```

chanwit (changes requested):

chanwit (approved):

PR #627, opened 2023-05-30
add renaming for varsFrom keys

### LINKED ISSUE #620: Implement Key Renaming Mechanism for Secret Vars in Terraform `varsFrom`
As part of our Terraform Controller Resource (CR), we would like to have the ability to use secret keys under different names. For example, we want the key `nodeCount` to be consumable as `node_count`. 

A possible approach is to specify the mapping in the `varsKeys` field of the Secret kind.

Here's an example:

```yaml
  varsFrom:
  - kind: Secret
    name: cluster-config
    varsKeys:
    - nodeCount:node_count
    - instanceType:instance_type
```

**User Story:**

As a Terraform Controller user, I want to be able to use secret keys under different names, so that I can customize key naming in line with my own naming conventions and requirements. For example, I want to consume `nodeCount` as `node_count` in my Terraform codes.

**Acceptance Criteria:**

- When the Terraform Controller user specifies a Secret with a key-value pair in the `varsKeys` field, the key should be mapped to a new name if specified.

- The user should be able to consume the renamed key in their Terraform configuration.

- The system should handle both cases where a renaming is specified (e.g. `nodeCount:node_count`) and where it is not specified (just `nodeCount`).

- Proper validation should be done to check if the renaming doesn't conflict with existing keys or reserved names.

- If a key doesn't exist in the Secret or there is any other issue with the renaming process, appropriate error handling and messages should be in place.

chanwit:
Fixes #620

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -21,7 +21,7 @@ k8s_yaml(
 k8s_kind('Terraform', image_json_path='{.spec.runnerPodTemplate.spec.image}', pod_readiness='ignore')
 
 k8s_resource(
-  objects=['helloworld:GitRepository:flux-system'],
+  objects=['helloworld:GitRepository:flux-system','helloworld-tf:Secret:flux-system'],
   workload='helloworld-tf',
   extra_pod_selectors={'instance': 'helloworld-tf'},
   pod_readiness='ignore',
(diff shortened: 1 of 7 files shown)


yitsushi (approved):

Nalum (approved):
LGTM

PR #633, opened 2023-05-30
add host aliases to the runner pod template

### LINKED ISSUE #631: support `hostAliases` in runner pod template
Currently, we support only a small subset of Pod fields.
We need a capability to specify `hostAliases` in the runner pod template, like this:

```yaml
  hostAliases:
  - hostnames:
   - abc.kubernetes.docker.internal
   - def.kubernetes.docker.internal
   - ijk.kubernetes.docker.internal
```

chanwit:
Fixes #631


yiannistri (approved):
LGTM

Issue #576, opened 2023-04-24
[tracker] Start plan-only mode for Terraform CR when creating a new branch on GitHub


chanwit:
### User story

As a developer, I want to be able to create a new branch and make changes to the Terraform configuration files in a "plan-only" mode so that I can review the plan information before merging the changes into the main branch.

### Subtasks:
 - [x] #577
 - [x] #578
 - [x] #580
 - [x] #581
 - [x] #582
 - [x] #583
 - [x] #605
 - [x] #588
 - [x] #628
 - [x] #670
 - [x] #661
 - [x] #662 X
 - [x] #663
 - [x] #586
 - [x] #664 X
 - [x] #669
 - [x] #665

### Acceptance criteria:
- [x] Successfully implement a feature flag for the branch-based planner system, allowing it to be toggled on or off.
- [x] Establish a clear communication mechanism between GitHub and the planner system, ensuring seamless integration.
- [x] Look at your information (PRs).
- [x] Design and create a mini controller capable of creating Terraform CR and its sourceRef for the planning mode by talking the K8s API server
- [x] Watching comments, enabling the correct triggering of actions based on the command in a comment.
- [ ] Develop and execute a comprehensive testing strategy to validate end-to-end communication between GitHub, the hook receiver, the mini controller, and the K8s API server, ensuring the functionality meets the desired requirements.
- [ ] User guide & documentations

### Success metric:
- [ ] Get feedback from one customer or prospect

chanwit:
ping @yitsushi @yiannistri

LappleApple:
Intermediary step: post in our Slack channel to drive feedback

Next step/milestone: Support more git providers, starting with GitLab with Azure DevOps

chanwit:
https://github.com/weaveworks/tf-controller/issues/691

PR #644, opened 2023-06-05
feat: polling mechanism to fetch and process Pull Requests

yitsushi:
## Git provider layer

This module aims to be a proxy between the codebase and different libraries to handle different providers. For now, only GitHub is implemented with 'jenkins-x/go-scm', but different providers can use different libraries under the hood if given feature is not available with go-scm. For example if we want to support Azure with comments, as I see it's not implemented for Azure[^1] yet, but it is possible we can find a library that does it, or we can implement our own on top go-scm like we did for committing multiple files in WGE[^2] because go-scm did not support multiple files in one commit.

[^1]: https://github.com/jenkins-x/go-scm/blob/1ec144b214fb9ac978b9260caa0483c05f94b93f/scm/driver/azure/issue.go#L81
[^2]: https://github.com/weaveworks/weave-gitops-enterprise/blob/main/pkg/git/azure.go#L190


Tiltfile
```diff
@@ -8,6 +8,12 @@ docker_build(
     dockerfile='Dockerfile',
 )
 
+docker_build(
+    'weaveworks/branch-based-planner',
+    context='.',
+    dockerfile='planner.Dockerfile',
```
yiannistri, line 14: This should `branch-based-planner.dockerfile` right?

yitsushi:
Quick sketch about follow-up in mind how to split the work up so more people can work on it:

* Add BBP to Tilt.
* Handle List Pull Requests[^1].
* Add 2 functions and create a plan[^2].
  * Check pull request update.
  * Check comments under a PR.
* Execute the plan.
* Post comment from Informer on new Plan Output.

[^1]: I would say, now just list all existing BBP Terraform objects and check their state against PRs, no in-memory cache that can break on restart and it would replan a lot.
[^2]: With a plan if we have a `!restart` and a new commit, restart the plan only once.

yitsushi:
Created issues based on the outline above, all linked in https://github.com/weaveworks/tf-controller/issues/576

charts/tf-controller/values.yaml
```diff
@@ -171,3 +171,7 @@ metrics:
 # -- Branch Based Planner-specific configurations
 branchBasedPlanner:
   enabled: false
+  replicaCount: 1
```
yiannistri, line 174: What is the intended behaviour when running multiple replicas of the branch-based planner? Is there any locking in place that ensures that pull requests are not acted upon by multiple replicas?

Tiltfile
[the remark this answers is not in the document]
yitsushi: Yes, I mentioned Tilt does not work. The tf-controller itself is working and starts with `tilt up` (at least it did for me).

See: https://github.com/weaveworks/tf-controller/issues/661

charts/tf-controller/values.yaml
yitsushi: I don't know. I did not spend too much time on the Helm Chart, tried to fix mostly the image name. I can revert those changes completely if it's confusing.

Touched the Chart only because first I though Tilt uses the chart, but it does not use it, so it was useless to change for me.

Tiltfile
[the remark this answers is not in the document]
yitsushi: Removed Tilt and Chart changes.

charts/tf-controller/values.yaml
yitsushi: Removed Tilt and Chart changes.

internal/server/polling/server.go
```diff
@@ -0,0 +1,100 @@
...
+		case <-ctx.Done():
+			return nil
+		default:
+		}
+
+		// Read the config in each iteration. The idea behind this decision is to
+		// allow the user to change the list of resources without the need of
+		// restart of the pod.
+		// It can be a bit smarter like using a time.Ticker and refresh config
+		// preiodically.
```
yiannistri, line 49:
```suggestion
		// periodically.
```

internal/server/polling/config.go
```diff
@@ -0,0 +1,61 @@
...
+
+func (s *Server) readConfig(ctx context.Context) (*Config, error) {
+	configMap := &corev1.ConfigMap{}
+	err := s.clusterClient.Get(ctx, s.configMapRef, configMap)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get ConfigMap: %w", err)
+	}
+
+	config := &Config{}
+	config.SecretNamespace = configMap.Data["secretMamespace"]
```
yiannistri, line 51:
```suggestion
	config.SecretNamespace = configMap.Data["secretNamespace"]
```

internal/server/polling/config.go
```diff
@@ -0,0 +1,61 @@
...
+	config.SecretName = configMap.Data["secretMame"]
```
yiannistri, line 52:
```suggestion
	config.SecretName = configMap.Data["secretName"]
```

internal/server/polling/config.go
```diff
@@ -0,0 +1,61 @@
...
+//
+// The secret is as reference to a secret with a 'token' key.
+//
+// ---
+// kind: ConfigMap
+// metadata:
+//   name: branch-based-planner
+// data:
+//   # Secret to use to use GitHub API
+//   secretMamespace: flux-system
```
yiannistri, line 24:
```suggestion
//   secretNamespace: flux-system
```

cmd/branch-based-planner/flags.go
```diff
@@ -21,9 +23,13 @@ type applicationOptions struct {
...
 
-	flag.StringVar(&opts.serverAddr,
-		"bind-address", webhook.DefaultListenAddress,
-		"The address the webhook server endpoint binds to.")
+	flag.StringVar(&opts.pollingConfigMap,
+		"polling-configmap", polling.DefaultConfigMapName,
+		"Namespace and name of the ConfigMap for the polling service.")
+
+	flag.DurationVar(&opts.pollingInterval,
+		"polling-intervak", polling.DefaultPollingInterval,
```
yiannistri, line 31:
```suggestion
		"polling-interval", polling.DefaultPollingInterval,
```

internal/server/polling/config.go
```diff
@@ -0,0 +1,61 @@
...
+	"gopkg.in/yaml.v3"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const DefaultConfigMapName = "default/branch-based-planner"
+
+// Example ConfigMap
+//
+// The secret is as reference to a secret with a 'token' key.
```
yiannistri, line 16:
```suggestion
// The secret is a reference to a secret with a 'token' key.
```

internal/git/provider/repo.go
```diff
@@ -1,7 +1,17 @@
...
+
 type Repository struct {
 	Project string
 	Org     string
 	Name    string
 }
+
+func (r Repository) String() string {
+	if r.Project == "" {
+		return fmt.Sprintf("%s/%s", r.Org, r.Name)
```
yiannistri, line 13: Is this meant to support Azure DevOps repositories in the future?

internal/git/provider/provider.go
```diff
@@ -2,20 +2,75 @@ package provider
...
+	targetProvider := ""
+	repo := Repository{}
+
+	parsedURL, err := url.Parse(repoURL)
+	if err != nil {
+		return nil, repo, fmt.Errorf("unable to parse url: %w", err)
+	}
+
+	// That's a pretty naiv check for now.
+	// Use proper parsing later with a well tested library.
```
yiannistri, line 53: Let's add a new issue for this so that we don't forget 👍

internal/git/provider/github_provider.go
```diff
@@ -1,16 +1,111 @@
...
+
+	prs := []PullRequest{}
+
+	for _, pr := range prList {
+		prs = append(prs, PullRequest{
+			Repository: repo,
+			Number:     pr.Number,
+			BaseBranch: pr.Base.Ref,
+			HeadBranch: pr.Head.Ref,
+			HeadRef:    pr.Head.Ref,
```
yiannistri, line 34: Is this right? Both being assigned to the same value?

internal/server/polling/config.go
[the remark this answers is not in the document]
yitsushi: For now yes, single token for multi repos.

internal/git/provider/repo.go
yitsushi: Yes

internal/server/polling/config.go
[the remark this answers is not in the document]
yiannistri: Ok we should include this in our docs.

internal/git/provider/github_provider.go
yitsushi: Good catch

yitsushi: Checked and one can be removed. I think originally I Ref in both cases, but renamed to Branch and left it there to rename it to Sha and populate with that value. So I did that instead of removing it.

internal/git/provider/provider.go
yitsushi: https://github.com/weaveworks/tf-controller/issues/669

yitsushi:
```
{"level":"info","ts":"2023-06-14T16:19:15.770+0200","logger":"informer","msg":"Starting branch-based planner informer","version":"","sha":""}
{"level":"info","ts":"2023-06-14T16:19:15.771+0200","logger":"polling-server","msg":"Starting polling server","version":"","sha":""}
{"level":"info","ts":"2023-06-14T16:19:16.366+0200","logger":"polling-server","msg":"pull request","version":"","sha":"","pr":{"Repository":{"Project":"","Org":"yitsushi","Name":"tf-controller-helloworld"},"Number":1,"BaseBranch":"main","HeadBranch":"yitsushi-patch-1","BaseSha":"a570bd81090f78cf9276dc61a0f0c1ce59d56df4","HeadSha":"6cb1de8b3a9a522388f888084506acef9d9e5a04"}}
```

internal/git/provider/github_provider.go
```diff
@@ -1,16 +1,112 @@
...
+
+	return nil
+}
+
+func (p *GitHubProvider) SetToken(tokenType, token string) error {
+	switch tokenType {
+	case APITokenType:
+		p.apiToken = token
+	default:
+		return fmt.Errorf("uknown token type: %s", tokenType)
```
yiannistri, line 67:
```suggestion
		return fmt.Errorf("unknown token type: %s", tokenType)
```

internal/server/polling/config.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const DefaultConfigMapName = "default/branch-based-planner"
+
+// Example ConfigMap
+//
+// The secret is a reference to a secret with a 'token' key.
+//
+// ---
+// kind: ConfigMap
```
yiannistri, line 19:
```suggestion
// apiVersion: v1
// kind: ConfigMap
```

yiannistri (approved):
Tested locally and it polls new PRs sucessfully 👌

yiannistri (approved):
LGTM 👍

internal/server/polling/server.go
```diff
@@ -0,0 +1,107 @@
...
+		time.Sleep(s.pollingInterval)
+	}
+}
+
+func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret *corev1.Secret) error {
+	if secret == nil {
+		return fmt.Errorf("secret is not defined")
+	}
+
+	tf, err := s.getTerraform(ctx, resource)
```
squaremo, line 78: I'm not sure that getTerraform and getSecret are doing much for you here, they don't add anything other than declaring the object for you. Is there any prospect of getting these things any way other than via the client? I think you can just save a couple of func definitions (and future readers looking them up) and do

```
var tf Terraform
err := s.clusterClient.Get(ctx, resource, &tf)
```

(Side note: at present, if a secret or terraform isn't there, you'll get an error something like `failed to get Terraform object: unable to get Terraform: not found`, which repeats itself)

internal/git/provider/github_provider.go
```diff
@@ -1,16 +1,112 @@
...
+	"github.com/go-logr/logr"
+	"github.com/jenkins-x/go-scm/scm"
+	"github.com/jenkins-x/go-scm/scm/driver/github"
+	"github.com/jenkins-x/go-scm/scm/factory"
+	"golang.org/x/net/context"
+)
 
-func (p GitHubProvider) ListPullRequests(ctx context.Context, repo Repository) (_ []PullRequest) {
-	panic("not implemented") // TODO: Implement
+type GitHubProvider struct {
```
squaremo, line 13: Does this need to be exported -- I would think the idea is to go through `New`, and not to see the implementations directly, is that right?

internal/git/provider/provider.go
```diff
@@ -2,20 +2,79 @@ package provider
...
-	ListPullRequests(ctx context.Context, repo Repository) []PullRequest
-	AddCommentToPullREquest(ctx context.Context, repo PullRequest, comment []byte)
+	ListPullRequests(ctx context.Context, repo Repository) ([]PullRequest, error)
+	AddCommentToPullRequest(ctx context.Context, repo PullRequest, body []byte) (*Comment, error)
+
+	SetLogger(logr.Logger) error
+	SetToken(tokenType, token string) error
+	SetHostname(hostname string) error
+
+	Setup() error
```
squaremo, line 22: If these (Set* and Setup) are intended to be called only by `New` (and perhaps by tests in this package), they can remain unexported. Or put it this way: is it legit to call `New`, then call `Setup` on the result?

internal/informer/bbp/informer.go
```diff
@@ -105,14 +105,14 @@ func (i *Informer) updateHandler(oldObj, newObj interface{}) {
 
 	previous, ok := oldObj.(*tfv1alpha2.Terraform)
 	if !ok {
-		i.log.Info("received object is not a Terraform object")
+		i.log.Info("previous object is not a Terraform object", "object", oldObj)
```
squaremo, line 108: Is this expected? (I don't know the guts of informers that well). If not, this ought to be a warning or error log, or even a crash.

internal/informer/bbp/informer.go
squaremo, line 1: I saw this on scanning through --

```
if previous.Annotations[AnnotationKey] != AnnotationValue || current.Annotations[AnnotationKey] != AnnotationValue {
		i.log.Info("Terraform object is not managed by the branch-based planner")
		return
	}
```

Why do you care about the previous value, here?

internal/server/polling/config.go
```diff
@@ -0,0 +1,67 @@
...
+//
+// ---
+// apiVersion: v1
+// kind: ConfigMap
+// metadata:
+//   name: branch-based-planner
+// data:
+//   # Secret to use to use GitHub API.
+//   # Key in the secret: token
+//   secretNamespace: flux-system
```
squaremo, line 26: Is there a good reason to have a cross-namespace reference here? (Apologies if it's in a design doc somewhere and I missed it)
Usually, allowing cross-namespace references represents a vulnerability.

squaremo (commented):
This has been merged, but I'll post my comments anyway. There's only one thing I'd consider a serious problem, and that's the blocking in Server.Start().
Anyway, have a look and see if there's anything you want to follow up.

cmd/branch-based-planner/main.go
```diff
@@ -46,40 +36,27 @@ func main() {
 		NewLogger(opts.logOptions).
 		WithValues("version", BuildVersion, "sha", BuildSHA)
 
-	webhookCtx, webhookCancel := signal.NotifyContext(context.Background(), os.Interrupt)
+	pollingCtx, pollingCancel := signal.NotifyContext(context.Background(), os.Interrupt)
```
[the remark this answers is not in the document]
yitsushi: Today I don't know why I created two of them, back than I'm sure I had reasons to do that.

internal/server/polling/server.go
```diff
@@ -0,0 +1,107 @@
...
+		return fmt.Errorf("failed to get git provider: %w", err)
+	}
+
+	prs, err := gitProvider.ListPullRequests(ctx, repo)
+	if err != nil {
+		return fmt.Errorf("failed to list pull requests: %w", err)
+	}
+
+	for _, pr := range prs {
+		s.log.Info("pull request", "pr", pr)
```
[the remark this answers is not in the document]
yitsushi: yes, it's there only to see if it could list PRs and we have an issue to handle them.

internal/git/provider/github_provider.go
yitsushi: In most cases yes, that code is kind of 1:1 copy from enterprise and I try to keep them in a similar shape, so it's easier to fix things if we notice something in one repo, at least for the first few months (that's not an old thing yet).

internal/git/provider/provider.go
yitsushi: It is not intended to be call only from `New()`, it is called only from new, but not necessarily.

internal/informer/bbp/informer.go
yitsushi: It happens, that's why the log is there. I don't know why, but right now we did not care about it much because until this code is just a filler until we we have to work with it (like we have a terrafrom object and the plan runs with it).

cmd/branch-based-planner/main.go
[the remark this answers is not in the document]
squaremo: Concurrency is tricky :-) I think you can have one context, cancel from either loop. I don't think you need the latter loop to be in a goroutine, since there's nothing else you need to do. And you don't need to wait on ctx.Done() afterwards, since you'll only exit when the loop ends. (You could consider using a sync.WaitGroup if you want to make sure the goroutine has exited, before returning)

internal/server/polling/server.go
```diff
@@ -0,0 +1,107 @@
...
+const DefaultPollingInterval = time.Second * 30
+
+type Server struct {
+	log             logr.Logger
+	clusterClient   client.Client
+	configMapRef    client.ObjectKey
+	pollingInterval time.Duration
+}
+
+func New(options ...Option) (*Server, error) {
```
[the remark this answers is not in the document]
yitsushi: It can be, based on the provider. For example for github only Token is mandatory (i think), but other providers may require different type of token structure that needs a different function. For now only token would be a mandatory argument as a single string. Other providers may require different token that has 2 parts, this way we don't have to change the function signature in the future.

squaremo:
It'd be good to follow this up with some testing. I know it's a chore to get all the scaffolding together, but it'll be really useful later when more pieces come together (and it'll be harder to do from scratch!).

cmd/branch-based-planner/main.go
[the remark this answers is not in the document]
yitsushi: > I don't think you need the latter loop to be in a goroutine, 

That was a request to run both of them in a separate go routine, I don't know why.

internal/server/polling/server.go
```diff
@@ -0,0 +1,107 @@
...
+
+	source, err := s.getSource(ctx, tf)
+	if err != nil {
+		return fmt.Errorf("failed to get Source object: %w", err)
+	}
+
+	gitProvider, repo, err := provider.FromURL(
+		source.Spec.URL,
+		provider.WithLogger(s.log),
+		provider.WithToken("api-token", string(secret.Data["token"])),
```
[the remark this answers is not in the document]
yitsushi: yes and there is a [constant for that](https://github.com/weaveworks/tf-controller/blob/main/internal/git/provider/options.go#L7) 🤦‍♀️ 

Sadly, long running PRs has this very bad side effect, it can be a mess at the end.

internal/server/polling/config.go
squaremo: OK, sounds like it's provisional. It'd be good to record the things that are not intended to remain standing, maybe in comments and the PR description.

internal/informer/bbp/informer.go
squaremo: That sounds suspicious -- what do you see as the oldObj when you run it?

yitsushi:
```
{"level":"info","ts":"2023-06-15T12:40:21.432+0200","logger":"informer","msg":"previous object is not a Terraform object","version":"","sha":"","object":{"apiVersion":"infra.contrib.fluxcd.io/v1alpha2","kind":"Terraform","namespace":"flux-system","name":"helloworld"}}
```

It shouldn't throw an error, but after this the state is `ok == false && previous == nil`.

So yes sus, but we will deal with it [soon](https://github.com/weaveworks/tf-controller/issues/664).

internal/server/polling/terraform.go
```diff
@@ -0,0 +1,49 @@
...
+func (s *Server) getSource(ctx context.Context, tf *v1alpha2.Terraform) (*sourcev1b2.GitRepository, error) {
+	if tf.Spec.SourceRef.Kind != sourcev1b2.GitRepositoryKind {
+		return nil, fmt.Errorf("branch based planner does not support source kind: %s", tf.Spec.SourceRef.Kind)
+	}
+
+	ref := client.ObjectKey{
+		Namespace: tf.Spec.SourceRef.Namespace,
+		Name:      tf.Spec.SourceRef.Name,
+	}
+	obj := &sourcev1b2.GitRepository{}
```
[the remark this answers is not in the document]
yitsushi: No reason, just picked one that was available on my cluster. It should be a bit more complex check.

For comments, I planned to go through and comment on everything, but I tried to let it merged as soon as possible as it blocks all other open issues and some of them can be done without those extra information.

internal/server/polling/server.go
yitsushi: It adds an extra layer because we may have to change the query to handle if the source is v1, v2, alphasometingelse. Right now `Terraform` is a `v1alpha2` and `Source` is `v1beta2`, it can be `v1`. So in the future this function can grow, extracting it into a function ensures no one tries to add all those into one function and let it grow 80-100+ lines.

internal/server/polling/server.go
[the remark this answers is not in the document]
squaremo: :nods:

internal/server/polling/server.go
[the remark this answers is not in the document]
squaremo: This makes me think perhaps the API is not quite the right shape: it sounds like only some permutations of options will give you a valid object (on API design, I always go back to http://sweng.the-davies.net/Home/rustys-api-design-manifesto). But, I lack a constructive suggestion for improving it right now, so I'll drop this bone :-)

internal/informer/bbp/informer.go
yitsushi: That was one of my thoughts about it, it does not give me the object, it gives me the reference only. The example, I used to build up this, used it as a kubernetes object, so I assumed it is a kubernetes object. It is still possible we only see a serialized version of the object and that does not contains all information, at the end `oldObj` and `newObj` are only an `interface{}`.

PR #650, opened 2023-06-06
fix logging for tf.output

### LINKED ISSUE #649: Fix logging for tf.output
We should wrap tf.Output with logging disabled as well.
```
outputs, err := r.tf.Output(ctx) 
```

chanwit:
Fixes #649

This PR is required to backport to v0.14.4.

Note: Trivy scan will be addressed later when Alpine ships the new libssl3 / libcrypto3 packages.

### DIFF
--- a/runner/server_outputs.go
+++ b/runner/server_outputs.go
@@ -3,8 +3,10 @@ package runner
 import (
 	"context"
 	"fmt"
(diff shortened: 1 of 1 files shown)


yitsushi (approved):
lgtm

Issue #481, opened 2022-12-14
Termination grace period is not respected when the pod is killed.


TarasLykhenko:
Hi Team,
We started to notice weird behaviour where many locks were not released when the tf-runners were killed, even when the graceful shutdown was set to 1 hour. 
After some investigation, we noticed that when the pod was restarted, the SIGKILL forcefully killed the Terraform process; therefore, it doesn’t start a graceful shutdown, which sometimes ends with the state lock not being released.

**We found these related issues:**
https://github.com/hashicorp/terraform-exec/pull/332
https://github.com/hashicorp/terraform-exec/pull/334

logs:
```
2022-11-29T21:42:10.266800268Z		Error message: workspace already locked (lock ID:
2022-11-29T21:42:10.266795895Z		
2022-11-29T21:42:10.266789968Z		Error: Error acquiring the state lock
2022-11-29T21:42:10.266760286Z		
2022-11-29T21:42:07.606685256Z	info	creating a plan
2022-11-29T21:42:07.587542369Z	info	mapping the Spec.VarsFrom
2022-11-29T21:42:07.587537646Z	info	mapping the Spec.Vars
2022-11-29T21:42:07.587512719Z	info	setting up the input variables
2022-11-29T21:42:07.58182462Z		Terraform has been successfully initialized!
2022-11-29T21:42:07.581792524Z		
2022-11-29T21:42:07.581420214Z		you run "terraform init" in the future.
2022-11-29T21:42:07.581416395Z		so that Terraform can guarantee to make the same selections by default when
2022-11-29T21:42:07.58141258Z		selections it made above. Include this file in your version control repository
2022-11-29T21:42:07.581408747Z		Terraform has created a lock file .terraform.lock.hcl to record the provider
2022-11-29T21:42:07.581401961Z		
2022-11-29T21:42:00.820815414Z		Initializing provider plugins...
2022-11-29T21:42:00.820797552Z		
2022-11-29T21:42:00.481037098Z		use this backend unless the backend configuration changes.
2022-11-29T21:42:00.481032663Z		Successfully configured the backend "remote"! Terraform will automatically
2022-11-29T21:42:00.48101017Z		
2022-11-29T21:42:00.198431966Z		Initializing the backend...
2022-11-29T21:41:58.413545513Z		Upgrading modules...
2022-11-29T21:41:58.386167971Z		}
2022-11-29T21:41:58.38616475Z		  "terraform_outdated": true
[...]

PR #667, opened 2023-06-13
honor pod grace period seconds in case of controller restart

### LINKED ISSUE #531: Restarting tf-controller kills tf-runner pods without graceful shutdown
# Issue description

We have observed that, if tf-controller is restarted for any reason whatsoever, after some delay the running tf-runner pods will be killed and not havea chance to terminate gracefully. 

This can be reproduced with a simple command:
```
kubectl rollout restart deploy/tf-controller -n flux-system
```

A result of this, in our case, are unreleased state file locks (we use s3 backend with dynamodb), in some cases corrupted state, and stray resources not added to the state. Overall, an absolute disaster.

### tf-runner doesn't handle SIGTERM properly

From what we see, whenever tf-runner receives SIGTERM, it doesn't pass it to `terraform` process, therefore not allowing it to gracefully shutdown/unlock state/etc. This is the cause for unreleased state locks and corrupted state.

I managed to hack a botched workaround for this: build my own runner image, and have my own shell script entrypoint that catches signals and `kill`s terraform process whenever SIGTERM is caught. This seemed to help, at least for this specific problem. However, instead of such workarounds I hope that tf-runner process can handle this task itself, as it should.

### random runner kills cause inconsistent state

Even if we solve the problem with SIGTERM, we're still left with another serious problem of having unnecessary runner kills that may cause inconsistent state. For example, whenever terraform creates a resource that takes a long time to provision, and waits for it to be created, it will only remember its ownership in the state after it was created completely. If during the wait period the process is interrupted, the likely scenario is that we don't have state record of a resource, but in reality resource already exists. All following apply attempts will throw "resource already exists" error, and only manual intervention (calling `terraform import`) can remediate the situation. Our declarative gitops approach is effectively broken at this point.

## Final thoughts

This is a huge problem, probably the biggest one we've encountered with tf-controller yet.
Pods in k8s are more-or-less ephemeral, and should be expected to terminate at any moment (e.g. during cluster scaledown event). Given terraform architecture with its reliance on state, it's probably impossible to fully prevent the problem. However, at the very least tf-controller shouldn't actively cause it be killing pods itself.

## Versions

```
tf-controller v0.13.1
```

### LINKED ISSUE #659: Repackage the container images when `libssl3` and `libcryto3` available on Alpine 3.18

chanwit:
Fixes #531 #481 
Also, Fixes #659 

Fix per observation from #531.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,11 +34,11 @@ LABEL org.opencontainers.image.source="https://github.com/weaveworks/tf-controll
 
 RUN apk update
 
-RUN apk add --no-cache ca-certificates tini git openssh-client gnupg && \
+RUN apk add --no-cache libcrypto3=3.1.1-r1 && \
+    apk add --no-cache libssl3=3.1.1-r1 && \
+    apk add --no-cache ca-certificates tini git openssh-client gnupg && \
     apk add --no-cache libretls && \
-    apk add --no-cache busybox && \
-    apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/main -U libcrypto3 && \
-    apk add --no-cache -X http://dl-cdn.alpinelinux.org/alpine/edge/main -U libssl3
+    apk add --no-cache busybox
 
 COPY --from=builder /workspace/tf-controller /usr/local/bin/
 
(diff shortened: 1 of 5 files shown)


Nalum (approved):
LGTM!

yiannistri:
Reviewing now 👀

chanwit:
Confirmed the behavior. Runner pods now correctly wait for the specified seconds before terminating.
For long running task, like cluster provisioning, we might need to adjust the value to be a bit large.

PR #674, opened 2023-06-16
Instructions for running the branch-based-planner

squaremo:
This adds a README under cmd/branch-based-planner, telling you how to set things up so it'll run and give some result.


yiannistri (approved):
Thanks for adding these instructions!🥇

yitsushi (approved):

PR #676, opened 2023-06-16
Add --no-cross-namespace-refs to branch-based-planner

squaremo:
The standard for Flux-source-integrating controllers is to include a flag `--no-cross-namespace-refs` for disabling cross-namespace access to sources. This commit adds the flag, and implements the ban when it's set.

This also fixes a bug whereby the namespace of the sourceRef was not defaulted to the namespace of the Terraform object containing it; so, if you left it out, the client.Get() for the source would fail.

### DIFF
--- a/cmd/branch-planner/flags.go
+++ b/cmd/branch-planner/flags.go
@@ -4,8 +4,10 @@ import (
 	"os"
 	"time"
 
+	"github.com/fluxcd/pkg/runtime/acl"
 	"github.com/fluxcd/pkg/runtime/logger"
 	flag "github.com/spf13/pflag"
+
 	"github.com/weaveworks/tf-controller/internal/server/polling"
 )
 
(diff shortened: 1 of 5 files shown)


yitsushi (commented):
I think we want the same check when we read the ConfigMap[^1][^2] and the Secret[^3] for the token. Is it planned in a separate PR?

[^1]: https://github.com/weaveworks/tf-controller/blob/main/internal/server/polling/config.go#L12
[^2]: https://github.com/weaveworks/tf-controller/blob/main/internal/server/polling/server.go#L60
[^3]: https://github.com/weaveworks/tf-controller/blob/main/internal/server/polling/server.go#L65

squaremo:
> Is it planned in a separate PR?

Yes that was my rough plan, but I'm not against doing it here if that's expedient.

yitsushi (approved):

PR #679, opened 2023-06-20
feat: add PlanOnly mode

### LINKED ISSUE #670: Plan only mode
Acceptance criteria:
* There is a `planOnly` field on the Terraform object.
* If `planOnly: true`, save the plan, but do not apply the changes, only plan.

yitsushi:
If `planOnly` is `true`, never apply terraform changes.

closes #670

### DIFF
--- a/controllers/tc000044_plan_only_mode_test.go
+++ b/controllers/tc000044_plan_only_mode_test.go
@@ -2,7 +2,6 @@ package controllers
 
 import (
 	"context"
-	"strings"
 	"testing"
 	"time"
 
@@ -134,22 +133,11 @@ func Test_000044_plan_only_mode_test(t *testing.T) {
 	}, timeout, interval).Should(Equal(1))
 
 	By("checking that the output ConfigMap contains the correct output data, provisioned by the TF resource.")
-	expectedOutputValue := map[string]string{
-		"Name":        "tfplan-default-" + terraformName,
-		"Namespace":   "flux-system",
-		"Value":       "Changes to Outputs:  + hello_world = \"Hello, World!\"You can apply this plan to save these new output values to the Terraformstate, without changing any real infrastructure.",
-		"OwnerRef[0]": string(createdHelloWorldTF.UID),
-	}
-	g.Eventually(func() (map[string]string, error) {
-		err := k8sClient.Get(ctx, outputKey, &planOutput)
-		value := string(planOutput.Data["tfplan"])
-		return map[string]string{
-			"Name":        planOutput.Name,
-			"Namespace":   planOutput.Namespace,
-			"Value":       strings.ReplaceAll(value, "\n", ""),
-			"OwnerRef[0]": string(planOutput.OwnerReferences[0].UID),
-		}, err
-	}, timeout, interval).Should(Equal(expectedOutputValue), "expected output %v", expectedOutputValue)
+
+	g.Expect(planOutput.Name).To(Equal("tfplan-default-" + terraformName))
+	g.Expect(planOutput.Namespace).To(Equal("flux-system"))
+	g.Expect(string(planOutput.OwnerReferences[0].UID)).To(Equal(string(createdHelloWorldTF.UID)))
+	g.Expect(string(planOutput.Data["tfplan"])).To(ContainSubstring(`+ hello_world = "Hello, World!"`))
 
 	It("should be stopped.")
 	By("checking the ready condition is still Plan within 5 seconds.")
--- a/controllers/tc000044_plan_only_mode_test.go
+++ b/controllers/tc000044_plan_only_mode_test.go
@@ -145,6 +145,7 @@ func Test_000044_plan_only_mode_test(t *testing.T) {
 					"Type":    c.Type,
 					"Reason":  c.Reason,
 					"Message": c.Message,
+					"Status":  c.Status,
 				}
 			}
 		}
@@ -153,6 +154,6 @@ func Test_000044_plan_only_mode_test(t *testing.T) {
 		"Type":    "Ready",
 		"Reason":  "TerraformPlannedWithChanges",
 		"Message": "Plan generated: set approvePlan: \"plan-main-b8e362c206\" to approve this plan.",
+		"Status":  metav1.ConditionTrue,
 	}))
-
 }
--- a/controllers/tc000044_plan_only_mode_test.go
+++ b/controllers/tc000044_plan_only_mode_test.go
@@ -111,14 +111,7 @@ func Test_000044_plan_only_mode_test(t *testing.T) {
 	By("checking that the TF resource existed inside the cluster.")
 	helloWorldTFKey := types.NamespacedName{Namespace: "flux-system", Name: terraformName}
 	createdHelloWorldTF := infrav1.Terraform{}
-	// We'll need to retry getting this newly created Terraform, Given that creation may not immediately happen.
-	g.Eventually(func() bool {
-		err := k8sClient.Get(ctx, helloWorldTFKey, &createdHelloWorldTF)
-		if err != nil {
-			return false
-		}
-		return true
-	}, timeout, interval).Should(BeTrue())
+	g.Expect(k8sClient.Get(ctx, helloWorldTFKey, &createdHelloWorldTF)).To(Succeed())
 
 	It("should be reconciled and produce the correct output secret.")
 	By("checking that the named output secret contains all outputs.")
--- a/controllers/tc000044_plan_only_mode_test.go
+++ b/controllers/tc000044_plan_only_mode_test.go
@@ -8,9 +8,11 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gstruct"
 
 	sourcev1 "github.com/fluxcd/source-controller/api/v1"
 	infrav1 "github.com/weaveworks/tf-controller/api/v1alpha2"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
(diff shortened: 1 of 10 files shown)


yitsushi:
I guess all the docs changes are here because it was not generated when those fields were added/changed and now when I run to generate, it generated it.

chanwit:
> I guess all the docs changes are here because it was not generated when those fields were added/changed and now when I run to generate, it generated it.

that's fine.

squaremo:
> I guess all the docs changes are here because it was not generated when those fields were added/changed and now when I run to generate, it generated it.

If you want to make life a bit easier for reviewers, you could generate with the new tool version _but not your changes_, then make your changes and generate again, in separate commits.

squaremo (approved):
I'm happy that this does what it says, and the test is enough to demonstrate so. So, :+1: and  thank you @yitsushi.

I think the test could be simplified, and I'd like to see incidental changes put in a separate PR (it's a good habit, because it makes reviewing easier), so please consider addressing those comments, with a reply or amendment, before merging.

controllers/tc000044_plan_only_mode_test.go
```diff
@@ -0,0 +1,176 @@
...
+	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+	infrav1 "github.com/weaveworks/tf-controller/api/v1alpha2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// +kubebuilder:docs-gen:collapse=Imports
+
+func Test_000044_plan_only_mode_test(t *testing.T) {
+	Spec("This spec describes the behaviour of a Terraform resource when PlanOnly is abled")
```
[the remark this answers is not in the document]
yitsushi: `abled` is a typo here, I wanted to write `abled` :woman_facepalming:

controllers/tc000044_plan_only_mode_test.go
```diff
@@ -0,0 +1,176 @@
...
+		},
+	}
+	It("should be created and attached successfully.")
+	g.Expect(k8sClient.Create(ctx, &helloWorldTF)).Should(Succeed())
+	defer func() { g.Expect(k8sClient.Delete(ctx, &helloWorldTF)).Should(Succeed()) }()
+
+	By("checking that the TF resource existed inside the cluster.")
+	helloWorldTFKey := types.NamespacedName{Namespace: "flux-system", Name: terraformName}
+	createdHelloWorldTF := infrav1.Terraform{}
+	// We'll need to retry getting this newly created Terraform, Given that creation may not immediately happen.
```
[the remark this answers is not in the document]
yitsushi: I coped most of the test from other cases to be sure I don't face the same issues others may did writing a lot of other test cases. I assume it has a reason why it's there.

chanwit:
Thank you for the PR @yitsushi and thank you for reviewing @squaremo 

I'll test this mode and get back to you shortly @yitsushi

chanwit:
```
❯ tfctl get
NAMESPACE       NAME            READY   MESSAGE         PLAN PENDING    AGE       
flux-system     helloworld-tf   Unknown Plan generated  true            4 minutes
```
- Message saying "Plan generated": correct.
- I was able to view plan via the `tfctl`: correct.

However, the Ready state would be "Ready" instead of "Unknown" - as we intended to make this object `green` in the planOnly mode.

chanwit:
```
❯ tfctl get
NAMESPACE       NAME            READY   MESSAGE         PLAN PENDING    AGE       
flux-system     helloworld-tf   Unknown Initializing    true            8 minutes
```

It's a bit tricky now. After the reconciliation kicked off next round, the object turned Unknow + Initializing.

controllers/tc000044_plan_only_mode_test.go
```diff
@@ -0,0 +1,177 @@
...
+	}, timeout, interval).Should(BeTrue())
+
+	It("should be reconciled and produce the correct output secret.")
+	By("checking that the named output secret contains all outputs.")
+	outputKey := types.NamespacedName{Namespace: "flux-system", Name: "tfplan-default-" + terraformName}
+	planOutput := corev1.ConfigMap{}
+	g.Eventually(func() (int, error) {
+		err := k8sClient.Get(ctx, outputKey, &planOutput)
+		if err != nil {
+			return -1, err
```
squaremo, line 131: You already checked that it exists, above; if there's an error now, that would represent an unexpected outcome, i.e,. a failure. I think it's better to use `g.Expect(k8sClient.Get(...)).To(Succeed())` here.

You can just remove this `Eventually` call, and change the assertion below. What you want to check is "the object exists, and as soon as it exists, it has this content". The above call to Eventually checks it exists, and the below can directly assert the content of `planObject` -- there's no further things to wait for that will change the result.

controllers/tc000044_plan_only_mode_test.go
```diff
@@ -0,0 +1,177 @@
...
+	}, timeout, interval).Should(Equal(1))
+
+	By("checking that the output ConfigMap contains the correct output data, provisioned by the TF resource.")
+	expectedOutputValue := map[string]string{
+		"Name":        "tfplan-default-" + terraformName,
+		"Namespace":   "flux-system",
+		"Value":       "Changes to Outputs:  + hello_world = \"Hello, World!\"You can apply this plan to save these new output values to the Terraformstate, without changing any real infrastructure.",
+		"OwnerRef[0]": string(createdHelloWorldTF.UID),
+	}
+	g.Eventually(func() (map[string]string, error) {
```
squaremo, line 143: As mentioned, this can just call `g.Expect` on the value of `planOutput`.

controllers/tf_controller_reconcile.go
```diff
@@ -283,8 +283,10 @@ func (r *TerraformReconciler) reconcile(ctx context.Context, runnerClient runner
 	if readyCondition != nil && readyCondition.Status == metav1.ConditionUnknown {
 		log.Info("Last known action was: " + lastKnownAction)
 		cond := terraform.Status.Conditions[readyConditionIndex]
-		if cond.Reason == infrav1.PlannedWithChangesReason && strings.HasPrefix(cond.Message, "Plan generated") {
-			// do nothing
+		if cond.Reason == infrav1.PlannedWithChangesReason && strings.HasPrefix(cond.Message, "Plan generated") && terraform.Spec.PlanOnly {
+			if terraform.Spec.PlanOnly {
```
squaremo, line 287: Is this already known to be `true` from the condition immediately above?

squaremo (commented):
Further alleged improvements :-)

controllers/tc000044_plan_only_mode_test.go
```diff
@@ -0,0 +1,177 @@
...
+			Digest:         "sha256:80ddfd18eb96f7d31cadc1a8a5171c6e2d95df3f6c23b0ed9cd8dddf6dba1406",
+			LastUpdateTime: metav1.Time{Time: updatedTime},
+		},
+	}
+
+	It("should be updated successfully.")
+	g.Expect(k8sClient.Status().Update(ctx, &testRepo)).Should(Succeed())
+
+	By("checking that the status and its URL gets reconciled.")
+	gitRepoKey := types.NamespacedName{Namespace: "flux-system", Name: sourceName}
```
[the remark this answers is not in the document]
yitsushi: Can we create a follow-up issue to clean up tests? I tried to keep tests similar to the rest as we have a lot (71 `_test.go` files each with one test case) so we don't get a a huge mess. When someone will add a new test case, it is most likely they will just copy an existing one and then that PR will have the same issues you mention here.

controllers/tc000044_plan_only_mode_test.go
```diff
@@ -0,0 +1,177 @@
...
+	createdRepo := &sourcev1.GitRepository{}
+	g.Expect(k8sClient.Get(ctx, gitRepoKey, createdRepo)).Should(Succeed())
```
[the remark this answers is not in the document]
yitsushi: (same as above)

controllers/tf_controller_reconcile.go
yitsushi: Good catch, the `&& terraform.Spec.PlanOnly` shouldn't be there.

controllers/tc000044_plan_only_mode_test.go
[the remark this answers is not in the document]
squaremo: We could stop making it worse by adding more of the same, today :-)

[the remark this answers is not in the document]
squaremo: (I did say originally all this was advisory, and I'll stick to that.)

controllers/tc000044_plan_only_mode_test.go
yitsushi: > You already checked that it exists, above;

Where? This is the first place where we get the `ConfigMap`. The `Eventually` is there to make sure it retries a few times and wait until the terraform controller creates the `ConfigMap`, it is created only after the plan output is available form the `terraform plan`.

This part can be moved out of the `Eventually` function:
```
return len(planOutput.Data), nil
```
But I don't think the `Eventually` can be removed (asked @chanwit, he may knows more if we need it or not).

squaremo: You are right, I got lost in the Eventuallys. You can discard the Eventually that follows and just assert the content of `planOutput`. (Unless I have misunderstood, and the plan object is created and populated later, by the controller).

chanwit:
It's not blocked and wait for manual intervention.

```
❯ tfctl get
NAMESPACE       NAME            READY   MESSAGE         PLAN PENDING    AGE       
flux-system     helloworld-tf   Unknown Initializing    true            2 minutes
```

controllers/tf_controller.go
squaremo, line 1: Would you explain what needed to be fixed, and what the fix is, in the commit message please.

chanwit (approved):

PR #683, opened 2023-06-20
docs: document ADR about polling, workflow, and responsibilities

### LINKED ISSUE #682: Add ADR about polling, workflow, and responsibilities

yitsushi:
Status will be updated when it's accepted.

Added an ADR to document the decision we use ADRs and linked some docs about it.

closes #682

### DIFF
--- a/docs/adr/0000-use-adrs-for-decisions.md
+++ b/docs/adr/0000-use-adrs-for-decisions.md
@@ -1,0 +1,32 @@
+# 1. Use ADRs to record decisions
+
+* Status: pending
+* Date: 2023-06-20
+* Authors: @yitsushi
+* Deciders: @yitsushi @chanwit @squaremo @yiannistri
+
+## Context
+
+Decisions that affect the development of Terraform Controller that are not
+captured via a proposal need to be captured in some way. We need a method that
+is lightweight and easy to discover the decision that have been made. The record
+of decisions will help future contributors to the project to understand why
+something has been implemented or is done a certain way.
+
+## Decision
+
+The project will use [Architectural Decision Records
+(ADR)](https://adr.github.io/) to record decisions that are made outside of a
+proposal.
+
+A [template](./0000-template.md) has been created based on prior work:
+
+* https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions
+* https://adr.github.io/madr/
+
+## Consequences
+
+When decisions are made that affect the entire project then a new ADR needs to
+be created. Likewise, if a decision has been superseded then we need to capture
+this as a new ADR and mark the previous ADR as superseded. Maintainers and
+contributors will need to decide when an ADR is to be created.
--- a/docs/adr/0000-use-adrs-for-decisions.md
+++ b/docs/adr/0000-use-adrs-for-decisions.md
@@ -1,6 +1,6 @@
 # 1. Use ADRs to record decisions
 
-* Status: pending
+* Status: proposed
 * Date: 2023-06-20
 * Authors: @yitsushi
 * Deciders: @yitsushi @chanwit @squaremo @yiannistri
--- a/docs/adr/0000-use-adrs-for-decisions.md
+++ b/docs/adr/0000-use-adrs-for-decisions.md
@@ -3,7 +3,7 @@
 * Status: proposed
 * Date: 2023-06-20
 * Authors: @yitsushi
-* Deciders: @yitsushi @chanwit @squaremo @yiannistri
+* Deciders: @yitsushi @chanwit @yiannistri
 
 ## Context
 
--- a/docs/adr/0001-pr-polling-workflow.md
+++ b/docs/adr/0001-pr-polling-workflow.md
@@ -1,4 +1,4 @@
-# 1. Pull Request Polling Workflow
+# 1. Pull Request Polling
 
 * Status: proposed
 * Date: 2023-06-20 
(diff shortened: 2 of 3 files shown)


docs/adr/0000-use-adrs-for-decisions.md
```diff
@@ -0,0 +1,32 @@
...
+of decisions will help future contributors to the project to understand why
+something has been implemented or is done a certain way.
+
+## Decision
+
+The project will use [Architectural Decision Records
+(ADR)](https://adr.github.io/) to record decisions that are made outside of a
+proposal.
+
+A [template](./0000-template.md) has been created based on prior work:
```
yiannistri, line 22: Should this file be included in this PR?

docs/adr/0001-pr-polling-workflow.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Context
+
+To detect pull request changes, we can use webhooks or polling using GitHub's
+API.
+
+## Decision
+
+We decided to use polling for security reasons. Using webhook would require to
+open an ingress to the cluster and that's not acceptable especially in an
```
squaremo, line 16: I don't find this universally-quantified statement very convincing. My estimation, coming from Flux, is that plenty of people are happy to use webhooks. I am fine with the _decision_, though. A better way to argue it would be that a significant proportion of people will not want to use webhooks, but polling will nearly always work, so it's better to start with polling.

docs/adr/0001-pr-polling-workflow.md
```diff
@@ -0,0 +1,55 @@
...
+air-gapped environment.
+
+The Branch-Based Planner has two components:
+
+1. Polling Server: Detect Pull Request changes and manage Teraform resource
+   state.
+2. Informer: Make a comment when new plan output is available.
+
+
+Full workflow and responsibilities:
```
squaremo, line 26: A lot of this is a design that would be relevant and accurate for webhooks, too. On the basis that an ADR should record one decision, perhaps this one should just explain the decision to use polling.

squaremo (commented):
I think you can cut the scope of the ADR down -- it doesn't need to cover the whole design, even as a sketch. I've suggested what I think would be a stronger argument in favour of polling over webhooks, too; I'll leave it to the other deciders to judge :-)

docs/adr/0001-pr-polling-workflow.md
yitsushi: > A better way to argue it would be that a significant proportion of people will not want to use webhooks

This is a decision record, it's not an RFC discussing where to go. It just records that we decided to use polling and the main reason was security and air-gapped environments.

docs/adr/0000-use-adrs-for-decisions.md
```diff
@@ -0,0 +1,32 @@
...
+
+* https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions
+* https://adr.github.io/madr/
+
+## Consequences
+
+When decisions are made that affect the entire project then a new ADR needs to
+be created. Likewise, if a decision has been superseded then we need to capture
+this as a new ADR and mark the previous ADR as superseded. Maintainers and
+contributors will need to decide when an ADR is to be created.
```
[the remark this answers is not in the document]
yitsushi: > who gets to be a decider?

That's not the point of the document. When an ADR is written we already know who was part of the decision.

> does an ADR need everyone to approve it?

That's a PR, like any other not an RFC. Review mostly for grammar, spelling, logical errors.

yitsushi:
> I'll leave it to the other deciders to judge :-)

The decision is not made during reviewing and ADR document PR. ADRs supposed to capture a decision made in the past. When we had more than one options to follow and we decided to use one or some of them and not the rest.

docs/adr/0000-use-adrs-for-decisions.md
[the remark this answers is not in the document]
yitsushi: Then I'll update that list, I though you were.

docs/adr/0001-pr-polling-workflow.md
squaremo: OK, so the correct place to register my objection to this line of reasoning is the RFC; I'll do that.

This makes a material difference, just in case that's in question -- if you read this rationale and took it at face value, you might think "we should never implement webhooks"; but I think that would be a poor conclusion to draw. Reasons matter.

docs/adr/0001-pr-polling-workflow.md
```diff
@@ -0,0 +1,30 @@
...
+* Deciders: @yitsushi @chanwit @yiannistri
+
+## Context
+
+To detect pull request changes, we can use webhooks or polling using GitHub's
+API.
+
+## Decision
+
+We decided to use polling for security reasons. Using webhook would require to
```
yiannistri, line 15:
```suggestion
We decided to start with polling for security reasons. Using webhooks would require users to
```

docs/adr/0001-pr-polling-workflow.md
```diff
@@ -0,0 +1,30 @@
...
+open an ingress to the cluster and that's not acceptable especially in an
+air-gapped environment.
```
yiannistri, line 17:
```suggestion
air-gapped environment. This does not mean that we will never consider using webhooks for this, but that initially, polling is what we have chosen to implement.
```

yiannistri (commented):
I have suggested a couple of edits to make it more obvious to readers, that webhooks may be supported in the future but for now we will build the first MVP using polling.

yiannistri (approved):
Thanks for capturing this @yitsushi 🥇

Issue #684, opened 2023-06-20
Add Enum type for storeReadablePlan in the docs


fire-ant:
Correctly document https://github.com/weaveworks/tf-controller/blob/ab48986d9d47060f5c96947a0951d43420104b96/api/v1alpha2/terraform_types.go#L237 
in https://weaveworks.github.io/tf-controller/References/terraform/#infra.contrib.fluxcd.io/v1alpha2.TerraformSpec

PR #685, opened 2023-06-20
update docs

fire-ant:
This fixes a discrepancy in the docs per #684 where the enum of the storeReadablePlan variable doesnt render in the documentation, leading the user to misunderstand what is available and applicable

### DIFF
--- a/cmd/branch-based-planner/README.md
+++ b/cmd/branch-based-planner/README.md
@@ -1,0 +1,94 @@
+# Branch-based planner development
+
+## How to run this
+
+You can run this with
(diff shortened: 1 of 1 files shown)


Nalum:
Looks good, just need to fix some logic in the code that checks the value: https://github.com/weaveworks/tf-controller/actions/runs/5324536107/jobs/9643972763?pr=685#step:13:590

chanwit:
@fire-ant could you please make sure that this is NOT a breaking change that might affect other products, like WGE?

PR #687, opened 2023-06-21
add finalizer to the dep only if object is not being deleted

### LINKED ISSUE #686: Unable to Add Finalizer During TF Object Deletion
While trying to perform a delete operation on a Terraform (TF) object, we're encountering an error related to adding a finalizer. The operation reports a failure due to the inability to add a finalizer to a Terraform object being deleted, with the following error message:

```
Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers ...
```

This issue often arises when running delete operations on TF objects and creates difficulties in managing these resources.

chanwit:
Fixes #686


yitsushi (approved):

chanwit:
Thank you @yitsushi !!

Issue #662, opened 2023-06-12
Handle the list of Pull Requests


yitsushi:
* Query existing branch-based planner Terraform objects.
* If there is no existing Terraform object create one (clone from the original one with changed fields: Terraform and Source).
* Save the ID of the PR and the latest commit hash on the Terraform object.

squaremo:
I think you will want to set an OwnerRef, too, so you can go back and find all the objects that _shouldn't_ exist (any more).

opudrovs:
@squaremo ok, thanks. Can you please elaborate a bit to what values (objects) the OwnerRef should be set to and on which objects?

For example, should it be set on the planner TF objects (each corresponding to a PR) to the original TF object (so that if that original object is deleted all planner TF objects cloned from the original TF object are garbage collected/deleted too)?

Or should it be set on the planner TF objects or GitRepositories corresponding to them and to which value?

squaremo:
> Can you please elaborate to what values (objects) the OwnerRef should be set to and on which objects?

A typical scheme for a reconciler loop is, given a "cause" object (in this case, the original Terraform object):

 1. Calculate all the objects that should exist (e.g., in this case a Terraform object and source for each PR on the relevant repo);
 1a. Set an owner reference on each, identifying the "cause" for its creation;
 2. Update or insert each of those objects as needed;
 3. Get a list of all objects that are owned by the cause object;
 4. Delete any of those that aren't part of the set calculated in 1.

Owner references have garbage collection implications that need to be carefully considered: https://kubernetes.io/docs/concepts/architecture/garbage-collection/ (these do roughly fit this situation: if the original Terraform object is deleted, you would want all its branch planning objects to also be deleted).

opudrovs:
Interesting, thank you!

Issue #586, opened 2023-05-03
Manage the state of a branch-based custom resource


yitsushi:
When receiving webhook events, create or delete custom resource related to the pull request.

As a starter:
* Create when a new Pull Request is opened
* Delete when the Pull Request is closed or merged.

More actions might be implemented in the future based on the outcome of #583

LappleApple:
Will make a follow-up issue to #662

PR #695, opened 2023-06-22
Handle the list of Pull Requests

opudrovs:
Closes https://github.com/weaveworks/tf-controller/issues/662
Closes https://github.com/weaveworks/tf-controller/issues/586

- Added the `createTerraform`, `createSource`, `createLabels`, and `createObjectName` methods.

- Added the `listTerraformObjects` method.

- Added the `deleteTerraform` and `deleteSource` methods.

- Reworked branch-based planner label keys and values.

- Added the `branchPollingInterval` flag.

- Added Terraform object and source reconciliation logic.

- Added the `reconcileTerraform` and `reconcileObjects` functions for Terraform object and source reconciliation.

- Added tests for creating, updating, and deleting objects during polling

- Added a new test target to run tests in the `internal` directory.

- Added running internal tests to the `test` workflow.

### DIFF
--- a/cmd/branch-based-planner/flags.go
+++ b/cmd/branch-based-planner/flags.go
@@ -10,8 +10,9 @@ import (
 )
 
 type applicationOptions struct {
-	pollingConfigMap string
-	pollingInterval  time.Duration
+	pollingConfigMap      string
+	pollingInterval       time.Duration
+	branchPollingInterval time.Duration
 
 	logOptions logger.Options
 
@@ -29,11 +30,19 @@ func parseFlags() *applicationOptions {
 
 	flag.DurationVar(&opts.pollingInterval,
 		"polling-interval", polling.DefaultPollingInterval,
-		"Wait between two request to the same Terraform object.")
+		"Wait between two requests to the same Terraform object.")
+
+	flag.DurationVar(&opts.branchPollingInterval,
+		"branch-polling-interval", 0,
+		"Interval to use for PR branch sources (default is to use the value of --polling-interval).")
 
 	opts.logOptions.BindFlags(flag.CommandLine)
 
 	flag.Parse()
+
+	if opts.branchPollingInterval == 0 {
+		opts.branchPollingInterval = opts.pollingInterval
+	}
 
 	opts.runtimeNamespace = os.Getenv("RUNTIME_NAMESPACE")
 
--- a/internal/informer/bbp/informer.go
+++ b/internal/informer/bbp/informer.go
@@ -90,25 +90,19 @@ func (i *Informer) SetDeleteHandler(fn func(interface{})) {
 }
 
 const (
-	AnnotationKey   = "terraform-conrtoller/branch-based-planner"
-	AnnotationValue = "true"
+	LabelKey            = "infra.weave.works/branch-based-planner"
+	LabelValue          = "true"
+	LabelPRIDKey string = "infra.weave.works/pr-id"
 )
 
 func (i *Informer) addHandler(obj interface{}) {}
 
-func (i *Informer) updateHandler(oldObj, newObj interface{}) {
+func (i *Informer) updateHandler(_, newObj interface{}) {
 	if !i.synced {
 		return
 	}
 	i.mux.RLock()
 	defer i.mux.RUnlock()
-
-	previous, ok := oldObj.(*tfv1alpha2.Terraform)
-	if !ok {
-		i.log.Info("previous object is not a Terraform object", "object", oldObj)
-
-		return
-	}
 
 	current, ok := newObj.(*tfv1alpha2.Terraform)
 	if !ok {
@@ -117,7 +111,7 @@ func (i *Informer) updateHandler(oldObj, newObj interface{}) {
 		return
 	}
 
-	if previous.Annotations[AnnotationKey] != AnnotationValue || current.Annotations[AnnotationKey] != AnnotationValue {
+	if current.Labels[LabelKey] != LabelValue {
 		i.log.Info("Terraform object is not managed by the branch-based planner")
 
 		return
--- a/internal/server/polling/server.go
+++ b/internal/server/polling/server.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/weaveworks/tf-controller/internal/git/provider"
+	"github.com/weaveworks/tf-controller/internal/informer/bbp"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -18,10 +19,11 @@ import (
 const DefaultPollingInterval = time.Second * 30
 
 type Server struct {
-	log             logr.Logger
-	clusterClient   client.Client
-	configMapRef    client.ObjectKey
-	pollingInterval time.Duration
+	log                   logr.Logger
+	clusterClient         client.Client
+	configMapRef          client.ObjectKey
+	pollingInterval       time.Duration
+	branchPollingInterval time.Duration
 }
 
 func New(options ...Option) (*Server, error) {
@@ -76,7 +78,7 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
 		return fmt.Errorf("secret is not defined")
 	}
 
-	tf, err := s.getTerraform(ctx, resource)
+	tf, err := s.getTerraformObject(ctx, resource)
 	if err != nil {
 		return fmt.Errorf("failed to get Terraform object: %w", err)
 	}
@@ -105,8 +107,34 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
 }
 
 func (s *Server) reconcile(ctx context.Context, original *infrav1.Terraform, source *sourcev1.GitRepository, prs []provider.PullRequest) error {
-	for _, pr := range prs {
-		s.log.Info("pull request", "pr", pr)
+	// List Terraform objects, created by the branch planner.
+	tfList, err := s.listTerraformObjects(ctx, original, map[string]string{
+		bbp.LabelKey: bbp.LabelValue,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to list Terraform objects: %w", err)
 	}
+
+	// Create and delete objects based on the diff
+	// between PRs and Terraform objects.
+	prMap := map[string]provider.PullRequest{}
+	for _, pr := range prs {
+		prMap[fmt.Sprintf("%d", pr.Number)] = pr
+
+		if err = s.reconcileTerraform(ctx, original, source, pr.HeadBranch, fmt.Sprintf("%d", pr.Number), s.branchPollingInterval); err != nil {
+			s.log.Error(err, "failed to reconcile Terraform object")
+		}
+	}
+
+	for _, branchTf := range tfList {
+		prLabel := branchTf.Labels[bbp.LabelPRIDKey]
+
+		if _, ok := prMap[prLabel]; !ok {
+			if err = s.deleteTerraform(ctx, branchTf); err != nil {
+				s.log.Error(err, "failed to delete Terraform object")
+			}
+		}
+	}
+
 	return nil
 }
--- a/internal/server/polling/terraform.go
+++ b/internal/server/polling/terraform.go
@@ -3,24 +3,48 @@ package polling
 import (
 	"context"
 	"fmt"
+	"time"
 
 	sourcev1b2 "github.com/fluxcd/source-controller/api/v1beta2"
-	"github.com/weaveworks/tf-controller/api/v1alpha2"
+	infrav1 "github.com/weaveworks/tf-controller/api/v1alpha2"
+	"github.com/weaveworks/tf-controller/internal/informer/bbp"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sLabels "k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-func (s *Server) getTerraform(ctx context.Context, ref client.ObjectKey) (*v1alpha2.Terraform, error) {
-	obj := &v1alpha2.Terraform{}
-	err := s.clusterClient.Get(ctx, ref, obj)
-	if err != nil {
+func (s *Server) getTerraformObject(ctx context.Context, ref client.ObjectKey) (*infrav1.Terraform, error) {
+	obj := &infrav1.Terraform{}
+	if err := s.clusterClient.Get(ctx, ref, obj); err != nil {
 		return nil, fmt.Errorf("unable to get Terraform: %w", err)
 	}
 
 	return obj, nil
 }
 
-func (s *Server) getSource(ctx context.Context, tf *v1alpha2.Terraform) (*sourcev1b2.GitRepository, error) {
+func (s *Server) listTerraformObjects(ctx context.Context, tf *infrav1.Terraform, labels map[string]string) ([]*infrav1.Terraform, error) {
+	tfList := &infrav1.TerraformList{}
+
+	if err := s.clusterClient.List(ctx, tfList,
+		client.MatchingLabelsSelector{
+			Selector: k8sLabels.Set(labels).AsSelector(),
+		},
+		client.InNamespace(tf.Namespace),
+	); err != nil {
+		return nil, fmt.Errorf("unable to list Terraform objects: %w", err)
+	}
+
+	result := make([]*infrav1.Terraform, len(tfList.Items))
+	for i := range tfList.Items {
+		result[i] = &tfList.Items[i]
+	}
+
+	return result, nil
+}
+
+func (s *Server) getSource(ctx context.Context, tf *infrav1.Terraform) (*sourcev1b2.GitRepository, error) {
 	if tf.Spec.SourceRef.Kind != sourcev1b2.GitRepositoryKind {
 		return nil, fmt.Errorf("branch based planner does not support source kind: %s", tf.Spec.SourceRef.Kind)
 	}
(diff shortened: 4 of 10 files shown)


internal/server/polling/server.go
```diff
@@ -97,9 +99,86 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
...
 	for _, pr := range prs {
 		s.log.Info("pull request", "pr", pr)
 	}
 
+	annotations := map[string]string{
+		bbp.AnnotationBBPKey: bbp.AnnotationBBPValue,
+	}
+
+	// List Terraform objects, created by BBP.
+	tfList, err := s.listTerraformObjects(ctx, tf, annotations)
```
squaremo, line 112: If you used a label for this purpose, you'd be able to use label selectors (the API server might even help out with indexing, and the controller-runtime cache certainly will).

squaremo: (You could consider using onwer references for this, too; but I understand if you want to restrict the scope of this PR).

internal/server/polling/server.go
```diff
@@ -97,9 +99,86 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
...
+	for _, pr := range prs {
+		shouldAddPR := true
+
+		for _, branchTF := range tfList.Items {
+			branchSource, err := s.getSource(ctx, &branchTF)
+			if err != nil {
+				continue
+			}
+
+			if pr.HeadBranch == branchSource.Spec.Reference.Branch {
```
squaremo, line 163: Isn't there an annotation for the PR being tracked? Are there any ways the test in this line could succeed when you don't want it to?

squaremo (commented):
Just comments for considering, since this is a draft.

internal/server/polling/server.go
opudrovs: @yitsushi is there a reason you want to use annotations instead of labels for defining which objects are managed by BBP?

internal/server/polling/server.go
opudrovs: @squaremo do you mean to use PR ID instead of branch name here? Can do it, thanks.

internal/server/polling/server.go
yitsushi: Nothing particular.

opudrovs: OK, then I can switch to using labels.

internal/server/polling/terraform.go
```diff
@@ -38,6 +61,114 @@ func (s *Server) getSource(ctx context.Context, tf *v1alpha2.Terraform) (*source
...
+	tf.SetLabels(s.createLabels(tf.Labels, branch, prID))
+
+	msg := fmt.Sprintf("Terraform object %s in the namespace %s", tfName, tf.ObjectMeta.Namespace)
+
+	source, err := s.createSource(ctx, originalSource, branch, prID)
+	if err != nil {
+		return fmt.Errorf("unable to create Source for %s: %w", msg, err)
+	}
+
+	tf.Spec.SourceRef.Name = source.ObjectMeta.Name
```
squaremo, line 84: This relies on the (safe) assumption that the copy of the source must be in the same namespace as the original source. A comment explaining why this assumption is safe would help people read the code -- or you could just defensively assign the namespace, so people don't even have to pause and think about it.

internal/server/polling/server.go
```diff
@@ -97,9 +99,75 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
...
+	// List Terraform objects, created by BBP.
+	tfList, err := s.listTerraformObjects(ctx, tf, labels)
+	if err != nil {
+		return fmt.Errorf("failed to list Terraform objects: %w", err)
+	}
+
+	// Create a list of objects to delete.
+	oldTFs := []*infrav1.Terraform{}
+
+	for _, branchTF := range tfList {
```
squaremo, line 121: It's not crucial here, but this and the following loop are quadratic -- they both, worst case, look through all the PRs x all the Terraform objects. Can we improve on that? (I am happy for this to be a follow up, since I think this is _correct_ as it stands).

squaremo (commented):
More comments while this is a draft. The only one I'd insist on (when this is ready for review) is the label name having a prefix particular to this controller; i.e., the prefix `infra.weave.works/`.

internal/informer/bbp/informer.go
```diff
@@ -117,7 +118,7 @@ func (i *Informer) updateHandler(oldObj, newObj interface{}) {
 		return
 	}
 
-	if previous.Annotations[AnnotationKey] != AnnotationValue || current.Annotations[AnnotationKey] != AnnotationValue {
+	if previous.Labels[LabelKey] != LabelValue || current.Labels[LabelKey] != LabelValue {
```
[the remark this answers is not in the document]
opudrovs: It is possibly not related to the scope of the current PR.

@yitsushi should I change this line in the current PR after the create/delete objects logic was added?

internal/server/polling/server.go
```diff
@@ -97,9 +99,75 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
 		return fmt.Errorf("failed to list pull requests: %w", err)
 	}
 
+	// Log the PRs.
 	for _, pr := range prs {
 		s.log.Info("pull request", "pr", pr)
 	}
 
+	labels := map[string]string{
```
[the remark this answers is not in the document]
opudrovs: Yes, thanks, I have not checked the PR yet, it's still in draft.

internal/server/polling/terraform.go
```diff
@@ -20,7 +22,28 @@ func (s *Server) getTerraform(ctx context.Context, ref client.ObjectKey) (*v1alp
...
+			Selector: k8sLabels.Set(labels).AsSelector(),
+		},
+		client.InNamespace(tf.ObjectMeta.Namespace),
+	); err != nil {
+		return nil, fmt.Errorf("unable to list Terraform objects: %w", err)
+	}
+
+	tfObjects := []*infrav1.Terraform{}
+
+	for _, obj := range tfList.Items {
```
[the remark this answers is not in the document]
opudrovs: The problem without copying was that when creating the slice of pointers only the last object was added on all positions of the slice.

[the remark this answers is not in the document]
opudrovs: Otherwise how can I convert a slice of TF objects (`Items`) to a slice of pointers to TF objects to return?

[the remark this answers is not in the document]
squaremo: `tfList.Items` is exactly the value you want, if you don't care about copying.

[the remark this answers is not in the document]
opudrovs: OK, thank you, got it!

```
result[i] = &tfList[i] // address of the entry
```

Here you mean `&tfList[i].Items`, correct?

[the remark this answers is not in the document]
squaremo: > Here you mean `&tfList[i].Items`, correct?

Yip! Good spot

internal/server/polling/terraform.go
opudrovs: What could be a reason not to put branch-based sources to the same namespace as the original source?

internal/server/polling/server.go
opudrovs: How exactly would you like to improve it?

internal/server/polling/terraform.go
```diff
@@ -20,7 +22,28 @@ func (s *Server) getTerraform(ctx context.Context, ref client.ObjectKey) (*v1alp
 	return obj, nil
 }
 
-func (s *Server) getSource(ctx context.Context, tf *v1alpha2.Terraform) (*sourcev1b2.GitRepository, error) {
+func (s *Server) listTerraformObjects(ctx context.Context, tf *infrav1.Terraform, labels map[string]string) ([]*infrav1.Terraform, error) {
+	tfList := &infrav1.TerraformList{}
+
+	if err := s.clusterClient.List(ctx, tfList,
+		client.MatchingLabelsSelector{
```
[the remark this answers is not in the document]
opudrovs: > More comments while this is a draft. The only one I'd insist on (when this is ready for review) is the label name having a prefix particular to this controller; i.e., the prefix `infra.weave.works/`

So, you would like to change prefix for both keys?

```
LabelKey            = "terraform-controller/branch-based-planner"
LabelPRIDKey string = "terraform-controller/pr-id"
```

from `terraform-controller/` to `infra.weave.works/`?

internal/informer/bbp/informer.go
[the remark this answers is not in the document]
yitsushi: I don't think we have to check the previous object. I don't even know if we have to check **anything** on the previous object, or we need it at all. We care about the generated `ConfigMap` with the plan linked to the `current` object, and doesn't really matter what was the previous state, we don't use values from there.

I added that check there mostly as a filler to prevent go from crying about "unused variable" and I didn't want to write `_ = previous` while tried to capture the fact we have that available.

Summary: I think we can even delete everything about the `previous` object as I don't see any reasons to keep it. Something like this:
```diff
-func (i *Informer) updateHandler(oldObj, newObj interface{}) {
+func (i *Informer) updateHandler(_, newObj interface{}) {
// ...
-	previous, ok := oldObj.(*tfv1alpha2.Terraform)
-	if !ok {
-		i.log.Info("previous object is not a Terraform object", "object", oldObj)
-
-		return
-	}
// ...
-	if previous.Annotations[AnnotationKey] != AnnotationValue || current.Annotations[AnnotationKey] != AnnotationValue {
+	if current.Annotations[AnnotationKey] != AnnotationValue {

```

internal/server/polling/server.go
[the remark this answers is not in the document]
opudrovs: @squaremo so, about the labels. Do you prefer to change it to smth. like that:

```
tfList, err := s.listTerraformObjects(ctx, tf, map[string]string{
    bbp.LabelKey: bbp.LabelValue,
})
```
or create the selector

```
client.MatchingLabelsSelector{
    Selector: k8sLabels.Set(labels).AsSelector(),
}
```

somewhere outside the `listTerraformObjects` method because it does not change and just pass it then to the `listTerraformObjects` function? And if the latter, where are we supposed to store such pre-created things, which are not constants? Like just create a "getter" method building and returning it?

internal/server/polling/server.go
```diff
@@ -97,9 +100,75 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
 		return fmt.Errorf("failed to list pull requests: %w", err)
 	}
 
+	// Log the PRs.
 	for _, pr := range prs {
 		s.log.Info("pull request", "pr", pr)
 	}
```
yitsushi, line 106: Don't forget to remove this.

internal/server/polling/server.go
```diff
@@ -97,9 +100,75 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
...
+		}
+
+		if shouldDeleteTF {
+			oldTFs = append(oldTFs, branchTF)
+		}
+	}
+
+	// Delete TF objects without corresponding PRs.
+	for _, oldTF := range oldTFs {
+		err = s.deleteTerraform(ctx, oldTF)
```
yitsushi, line 130: I think we can move it up into the previous double-loop. Right now we have a quadratic double-loop that allocates and constantly resize a new slice, just to loop over all items and call delete.

Even setting the capacity of `oldTFs` to `len(tfList)` can save us a lot of resizing. Or simply just get rid of this slice.

internal/server/polling/server.go
```diff
@@ -97,9 +100,75 @@ func (s *Server) poll(ctx context.Context, resource types.NamespacedName, secret
...
+			}
+		}
+
+		if shouldAddPR {
+			newPRs = append(newPRs, pr)
+		}
+	}
+
+	// Create Terraform objects for new PRs.
+	for _, pr := range newPRs {
```
yitsushi, line 157: Same as above.

internal/server/polling/server.go
yitsushi: I think as a first step, we can merge the two nested loops (delete tfs and create tfs).

For example:
```
// Start with PRs, they are the source of truth.
loop _, pr := prs {

  // Assume we have to create it.
  shouldCreate := true
  
  // Worst case, we iterated through all of them, that means we have to create one.
  loop idx, tf := tfList as tf {
    if pr belongs to tfList {
    
      // We don't care, there is a TF object for this PR, nothing to do.
      tfList[idx] := nil
      
      // We don't have to create it.
      shouldCreate = false
      
      // We can stop.
      break
    }
  }
  
  if shouldCreate {
    createResource(...)
  }
}

for _, tf := tfList {
  // It was not deleted from this list, so there is no related PR.
  if tf != nil {
    delete
  }
}
```

It still has that nested loop, but only once (and not twice as before) and we could remove two extra slices from the code (`oldTFs` and `newPRs`).

It may looks weird, but it can save a lot of cpu cycles and memory.

internal/server/polling/terraform.go
```diff
@@ -38,6 +62,119 @@ func (s *Server) getSource(ctx context.Context, tf *v1alpha2.Terraform) (*source
...
+
+	tf.ResourceVersion = ""
+
+	tfName := s.createObjectName(tf.ObjectMeta.Name, branch, prID)
+
+	tf.SetName(tfName)
+
+	tf.SetLabels(s.createLabels(tf.Labels, branch, prID))
+
+	msg := fmt.Sprintf("Terraform object %s in the namespace %s", tfName, tf.ObjectMeta.Namespace)
```
yitsushi, line 76: Can we group variable assignments and functions calls together? I like (and highly prefer) when the code has empty lines to break up logical sections, but not between every lines.

internal/server/polling/terraform.go
```diff
@@ -38,6 +62,119 @@ func (s *Server) getSource(ctx context.Context, tf *v1alpha2.Terraform) (*source
...
+	if err := s.clusterClient.Create(ctx, source); err != nil {
+		return nil, fmt.Errorf("unable to create %s: %w", msg, err)
+	}
+
+	s.log.Info(fmt.Sprintf("created %s", msg))
+
+	return source, nil
+}
+
+func (s *Server) createObjectName(name string, branch string, prID string) string {
```
yitsushi, line 128: [can be ignored]
It's fine here, but later we may want to move these helper functions out of the struct as the Informer can use it, maybe it will not use it at all, but I like taking extra notes.

opudrovs:
>I think as a first step, we can merge the two nested loops (delete tfs and create tfs).

I've already optimized the first quadratic loop, thinking about what t do with the second one. Will look into your code.

internal/server/polling/server.go
opudrovs: Still a draft PR. It's a leftover from your code. Log messages are there for convenience. Removed.

internal/server/polling/server.go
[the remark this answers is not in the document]
opudrovs: For now changed it to

```
tfList, err := s.listTerraformObjects(ctx, tf, map[string]string{
    bbp.LabelKey: bbp.LabelValue,
})
```

internal/server/polling/server.go
opudrovs: @yitsushi added your optimized loops.

opudrovs:
@squaremo @yitsushi 

What is preferred:

to use

```
SetName
GetName
SetNamespace
GetNamespace
```

and similar getters/setters for `Labels` and `Annotations`

vs

```
Name
Namespace
Labels
Annotations
```
to get and set values of these fields?

Are both ways functionally the same and does using getters/setters explicitly just prevent us from setting (`obj.GetName() = "blablabla"`) or assigning/returning (like `x := obj.SetName("blablabla")`) values accidentally, which in the examples will result in errors?

opudrovs:
In addition, is it ever recommended to use `obj.ObjectMeta.Name` instead of `obj.Name` in such cases?

Or should `ObjectMeta` be used only when initializing structures like

```
ObjectMeta: metav1.ObjectMeta{
    Name:      "blablaname",
    Namespace: "blablanamespace",
},
```
not when assigning values to individual fields like `obj.ObjectMeta.Name = "blablabla"`?

internal/server/polling/server.go
squaremo: If you use a map, you don't need the inner loop:

```go
prSet := map[string]provider.PullRequest
for _, pr := range prs {
    prSet[fmt.Sprintf("%d", pr.Number)] = pr
}

for _, branchTf := range tfList {
    prLabel := branchTf.GetLabels()[LabelPRIDKey]
    if pr, ok := prSet[prLabel]; !ok {
        // no PR for this TF; delete it
        s.deleteTerraform(ctx, branchTf)
    } else {
        // don't need to create a Terraform for this PR, one exists
        delete(prMap, prLabel)
    }
}
for _, pr := range prMap {
    s.createTerraform(ctx, tf, source, pr.HeadBranch, fmt.Sprintf("%d", pr.Number), s.branchPollingInterval); err != nil {
}
```

I left out handling the errors from `create-` and `deleteTerraform`, even so it's a bit shorter I think.

The set does double-duty by keeping track of the set of Terraform objects that should exist and don't. There's a subtle assumption here, which is that at most one Terraform object will have any given PR number (otherwise, all but one will be deleted).

squaremo: Two notes, though:

 1. It would be conventional to "upsert" all resources that are supposed to exist, since objects that already exist might nonetheless need their definitions updated. This simplifies things however you arrange loops. https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil has a func specifically for this.
 2. Having tests that demonstrate that it's correct would make it easier to change the particulars of the algorithm with confidence.

opudrovs: 1. What does upsert means in this context? Do you mean using
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate

function instead of client.Create?

2.
> Having tests that demonstrate that it's correct would make it easier to change the particulars of the algorithm with confidence.

Do you mean specific tests for upserting objects? Are they different from tests for creating or deleting objects we already discussed?

squaremo: By "upsert" I mean update if it needs updating (say if the original Terraform object had changed part of its spec), or create if it doesn't exist. This is really what the algorithm should do, I think, since you want to keep the branch Terraform objects up to date with respect to the original (and similarly the sources).

I mean tests that the algorithm does what it's supposed to: the equation is something like "given an original Terraform object and its source, for each pull request from the repo there should be a branch Terraform object the same as the original except with planOnly=true, and a source the same as the original source but referring to the pull request branch."

opudrovs: Ah, you mean update branch TF objects not based on the PRs (I think no changes should be required for this) but based on the original TF object which might have changed.

So, you would like to add updating branch TFs in a loop too based on the original TF object (like all fields at once with modifying some of them after the update?))? If yes, what function from that package should be used for this?

cmd/branch-based-planner/flags.go
```diff
@@ -31,10 +32,18 @@ func parseFlags() *applicationOptions {
 		"polling-interval", polling.DefaultPollingInterval,
 		"Wait between two request to the same Terraform object.")
 
+	flag.DurationVar(&opts.branchPollingInterval,
+		"branch-polling-interval", 0,
+		"Wait between two request to the same Terraform object.")
```
squaremo, line 37:
```suggestion
		"Interval to use for PR branch sources (default is to use the value of --polling-interval)")
```

opudrovs:
@squaremo @yitsushi so, I've added the `CreateOrUpdate` functions.

The main question:

What to do with setting `ResourceVersion`?

```
tf.SetResourceVersion("")
...
source.SetResourceVersion("")
```

If I don't add this, there is an error when creating objects:

```
{"level":"error","ts":"2023-06-28T20:56:25.758+0200","logger":"polling-server","msg":"failed to reconcile Terraform object","version":"","sha":"","error":"reconcile failed for Terraform object helloworld-tf-test-4-4 in the namespace default: resourceVersion should not be set on objects to be created"}
```

If I add this, the branch TF objects are created but are not updated after applying the modified YAML for the original object.

After updating the original YAML, I see a message that the object were updated but the fields I changed in the original TF are not updated:

```
{"level":"info","ts":"2023-06-28T21:13:22.576+0200","logger":"polling-server","msg":"Source helloworld-test-4-4 in the namespace default successfully reconciled","version":"","sha":"","operation":"unchanged"}

{"level":"info","ts":"2023-06-28T21:13:22.592+0200","logger":"polling-server","msg":"Terraform object helloworld-tf-test-4-4 in the namespace default successfully reconciled","version":"","sha":"","operation":"updated"}
```

If I got it right, the actual `ResourceVersion` is required for updates.

I tried to use the following check from the method docs

```
obj.ObjectMeta.CreationTimestamp.IsZero()
```

when setting the `ResourceVersion` to `""`,but it always equals false, outside or within the mutate functions.

opudrovs:
@squaremo @yitsushi   Additional questions:

2) Should these methods be called `createOrUpdateTerraform` and `createOrUpdateSource` (instead of create) or `reconcileTerraform` and `reconcileSource` (the latter might help to avoid confusion)? Because the messages in the example are speaking of reconciliation and our server method is called `reconcile`.

3) Is returning the source from `createOrUpdateSource` method OK? I am not sure it will be updated by that method (the original `CreateOrUpdate`), but we only need it to get its name and namespace for the TF spec.

4) I changed the verbs in the log messages on failure or successful creation/update from `create` or `update` to `reconcile`, similar to the example. Is it OK?

opudrovs:
I've added running tests in the `internal` directory and subdirectories to the `test` target. But as far as I can see, the `test` target is not run in any workflows, it can be run only locally. And I am getting a timeout for controller tests for some reason when running this target with added internal tests locally. Maybe this should be tacked in a separate issue.

yitsushi (approved):

squaremo (dismissed):
:100: for including tests, asking questions about the best way to do things, and otherwise for responding to the extended commentary :-)

May I request a couple of things before this is merged:
 - fix the `.PHONY` line (I've commented at the appropriate line)
 - update the commit summary and include a prose description of the changes in the commit message*.

(*here is some great advice on what to put in commit messages: https://cbea.ms/git-commit/#why-not-how)

yitsushi (approved):

PR #702, opened 2023-06-26
feat: build and publish images GitHub workflow

yitsushi:
Closes #697

PR #703, opened 2023-06-26
patch: don't build and publish images in the test workflow

### LINKED ISSUE #698: Avoid building Docker images in test workflow
Should follow after #697

The `test` workflow should not need to build Docker containers as that takes a long time.

yitsushi:
Closes #698

merge only after #702


yitsushi:
image
vs
image

yiannistri (approved):

PR #708, opened 2023-06-26
Adding git parser lib

### LINKED ISSUE #669: Use a well tested library to parse Git Repository URIs
Right now in https://github.com/weaveworks/tf-controller/pull/644 the git URI parsing is a simple one that satisfies the parsing of github repositories. 

See: https://github.com/weaveworks/tf-controller/pull/644#discussion_r1229142926

luizbafilho:
closes #669 

Adds externa lib to properly parse urls

### DIFF
--- a/.github/workflows/build-and-publish.yaml
+++ b/.github/workflows/build-and-publish.yaml
@@ -1,0 +1,116 @@
+name: Build and Publish Images
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+permissions:
+  id-token: write # needed for keyless signing
+  packages: write # needed for ghcr access
+
+env:
+  CONTROLLER: ${{ github.event.repository.name }}
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.19.x
+      - name: Setup Terraform
+        run: |
+          export TF_VERSION=1.3.7
+          wget https://releases.hashicorp.com/terraform/${TF_VERSION}/terraform_${TF_VERSION}_linux_amd64.zip
+          unzip -q terraform_${TF_VERSION}_linux_amd64.zip
+          mv terraform $(which terraform)
+          terraform --version
+      - name: Set up yq
+        uses: frenck/action-setup-yq@v1
+        with:
+          version: 4.14.1
+      - name: Setup Kustomize
+        uses: fluxcd/pkg/actions/kustomize@main
+      - name: Get branch names
+        id: branch-name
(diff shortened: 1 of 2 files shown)


internal/git/provider/provider.go
```diff
@@ -48,27 +51,17 @@ func FromURL(repoURL string, options ...ProviderOption) (Provider, Repository, e
...
-		repo.Org = parts[1]
-		repo.Name = strings.TrimSuffix(parts[2], ".git")
-	}
+	targetProvider = gitURL.GetProvider()
+	repo.Org = gitURL.GetOwnerName()
+	repo.Name = gitURL.GetRepoName()
 
-	if targetProvider == "" {
-		return nil, repo, fmt.Errorf("could not parse provider from url: %s", repoURL)
+	if targetProvider == AzureProviderName {
```
squaremo, line 63: The range of `GetProvider` is the consts in https://github.com/kubescape/go-git-url/blob/master/apis/provider.go, not the consts defined here. This comparison should really use the const from the library, otherwise it relies on the coincidence of the strings being the same.

internal/git/provider/provider_test.go
```diff
@@ -41,7 +41,7 @@ func TestFromURL(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
-		p, repo, err := provider.FromURL(testCase.url)
+		p, repo, err := provider.FromURL(testCase.url, provider.WithToken("api-token", "token"))
```
squaremo, line 44: Why is this necessary?

internal/git/provider/provider.go
```diff
@@ -48,27 +51,17 @@ func FromURL(repoURL string, options ...ProviderOption) (Provider, Repository, e
...
+		repo.Project = gitURL.(*azureparserv1.AzureURL).GetProjectName()
```
squaremo, line 64: I don't see a test covering this; I think it's worth adding one.

squaremo (approved):
This is nice and self-contained, thanks @luizbafilho. I've registered this as `Approve` but I think there's a couple of things that would be easy to change and worthwhile:

 - a test covering the special case for Azure
 - use the consts from the library in comparisons with return values from the library

internal/git/provider/provider_test.go
squaremo: OK; would you put it in a separate commit, and explain in the commit message what was broken and how this fixes it, so future readers understand why this change is here. Thank you!

internal/git/provider/provider.go
squaremo: That works, good thinking.

internal/git/provider/provider.go
squaremo: Do we need the special case then?

squaremo (approved):
:+1: thanks Luiz.
This is small enough that you could reasonably squash it (before, or in the process of, merging it).

PR #710, opened 2023-06-26
patch: fix force-unlock with delete path by updating in terraform.Status.Lock.Pending terraform CR.

### LINKED ISSUE #560: Force unlock doesn't work with delete path
When the Terraform CR is being deleted with 'destroyResourcesOnDeletion' set to true, it will delete the backend resources. If the tf-controller pod gets restarted during the backend resources deletion, the Terraform resource may get into a lock error state. To resolve the lock error, we are attempting to force an unlock with the following code:

```
 tfstate:
   forceUnlock: "yes"
   lockIdentifier: 49829de4-6475-d73f-1785-489ecb1c8e1d
```

However, it appears that this method does not work when the Terraform resource is being deleted. Therefore, we request support for force unlock with the delete path.

mmeha:
finalize() routine updates terraform.Status.Lock.Pending in terraform CR on lock error.
Return updated terraform from routine finalize() so that it can be patched on lock error.

closes #560

### DIFF
--- a/controllers/tf_controller.go
+++ b/controllers/tf_controller.go
@@ -391,7 +391,12 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	traceLog.Info("Check for deletion timestamp to finalize")
 	if !terraform.ObjectMeta.DeletionTimestamp.IsZero() {
 		traceLog.Info("Calling finalize function")
-		if result, err := r.finalize(ctx, terraform, runnerClient, sourceObj, reconciliationLoopID); err != nil {
+		if terraform, result, err := r.finalize(ctx, terraform, runnerClient, sourceObj, reconciliationLoopID); err != nil {
+			traceLog.Info("Patch the status of the Terraform resource")
+			if patchErr := r.patchStatus(ctx, req.NamespacedName, terraform.Status); err != nil {
+				log.Error(patchErr, "unable to update status after the finalize is complete")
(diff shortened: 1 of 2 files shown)


controllers/tf_controller.go
```diff
@@ -391,7 +391,12 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	traceLog.Info("Check for deletion timestamp to finalize")
 	if !terraform.ObjectMeta.DeletionTimestamp.IsZero() {
 		traceLog.Info("Calling finalize function")
-		if result, err := r.finalize(ctx, terraform, runnerClient, sourceObj, reconciliationLoopID); err != nil {
+		if terraform, result, err := r.finalize(ctx, terraform, runnerClient, sourceObj, reconciliationLoopID); err != nil {
+			traceLog.Info("Patch the status of the Terraform resource")
+			if patchErr := r.patchStatus(ctx, req.NamespacedName, terraform.Status); err != nil {
```
chanwit, line 396:
```suggestion
			if patchErr := r.patchStatus(ctx, req.NamespacedName, terraform.Status); patchErr != nil {
```

chanwit (changes requested):

chanwit:
Thank you for the PR @mmeha 
I'm reviewing it and will test it on a cluster soon.

chanwit (approved):

chanwit:
Thank you so much @mmeha :1st_place_medal:

PR #717, opened 2023-06-27
Supply var secret_subject to helloworld

squaremo:
https://github.com/tf-controller/helloworld/commit/7353dc3abb7db80acb1afc915686490fbb8ac714 added a variable to the helloworld main.tf used in these tests. Compensate by supplying a value for it in the Terraform objects.

### DIFF
--- a/config/testdata/approve-plan/test.yaml
+++ b/config/testdata/approve-plan/test.yaml
@@ -7,6 +7,9 @@ spec:
   interval: 10s
   approvePlan: "auto"
   path: ./
+  vars:
+  - name: secret_subject
+    value: shhhh
   sourceRef:
     kind: GitRepository
     name: helloworld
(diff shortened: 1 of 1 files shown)


squaremo:
(With this PR I mostly want to see if it'll fix the current e2e test failure, but it's quite possible other tests will still break. A better solution would be to give the new variable in helloworld a default, since that wouldn't require changes to tests.)

chanwit:
Sorry about that.

I reverted the code on the helloworld example.

squaremo:
> I reverted the code on the helloworld example.

Thanks Chanwit, that'll do it. I'll close this PR.

Interesting side note: the tests for this PR failed too, but in a slightly different way -- I changed the test Terraform objects so that they supply the variable secret_subject to the helloworld module, and it doesn't complain about that variable any more -- it complains about lacking the variable `subject` instead. That one has a default value, so I wouldn't expect to have to supply it (and it wasn't being supplied before). Maybe I don't understand how Terraform variables work!

https://github.com/weaveworks/tf-controller/actions/runs/5394903388/jobs/9796717630

PR #723, opened 2023-06-30
fix pending plan not equal plan id in secret

### LINKED ISSUE #599: Unknown desc = error pending plan and plan's name in  the secret are not matched
Hello,

 Last week I was starting several Terraform manifest inside GKE and I started received these messages `Unknown desc = error pending plan and plan's name in the secret are not matched: `

 This error has appeared in several Terraform, as you can see here:
Screenshot_20230515_154306

 Let me share more info just from 1 Terraform manifest, maybe i don't see what is the root of this issue:

- terraform 
[terraform.tar.gz](https://github.com/weaveworks/tf-controller/files/11478434/terraform.tar.gz)

- cm from plan 
[cm.tar.gz](https://github.com/weaveworks/tf-controller/files/11478439/cm.tar.gz)

- secret from plan 
[secret.tar.gz](https://github.com/weaveworks/tf-controller/files/11478436/secret.tar.gz)

Let me also describe my environment:
- kubernetes in Google, GKE:  Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.10-gke.2300"
- FluxCD v0.41.2
- Terraform Controller v0.14.0

Could you help me fixing this? 
Marc

### LINKED ISSUE #699: Allow to use short plan id to approve
Upgrading to Flux v2 is currently breaking the use of shot plan id to approve the plan.
We want this capability back.

Context: https://weaveworks.github.io/tf-controller/use_tf_controller/to_plan_and_manually_apply_Terraform_resources/

Without this fix, users need to supply the full plan name, which is annoying/long. 

**Acceptance Criteria:**
- [ ] short plan id works if you supply it (`plan-main-abcdef`).

chanwit:
Fixes #599
Fixes #699 

To cope with similar issues in the future, we'll use the short plan id everywhere.
- In the pending plan 
- In the last applied plan
- In the approval message
- Having a dedicated lib for converting a revision to a plan id, in `api/planid/plan_id.go` and having test cases to cover old and new formats.

### DIFF
--- a/api/planid/plan_id_test.go
+++ b/api/planid/plan_id_test.go
@@ -1,0 +1,52 @@
+package planid
+
+import (
+	"testing"
+)
+
+func TestGetPlanID(t *testing.T) {
+	tests := []struct {
+		name     string
+		revision string
+		want     string
+	}{
+		{
+			name:     "Valid hash more than 10 characters",
+			revision: "branch1@algo1:12345678901234567890",
+			want:     "plan-branch1-1234567890",
+		},
+		{
+			name:     "Valid hash equal to 10 characters",
+			revision: "branch2@algo2:1234567890",
+			want:     "plan-branch2-1234567890",
+		},
+		{
+			name:     "Valid hash less than 10 characters",
+			revision: "branch3@algo3:123456789",
+			want:     "plan-branch3-123456789",
+		},
+		{
+			name:     "Valid branch and hash",
+			revision: "branch4@algo4:abc123xyz",
+			want:     "plan-branch4-abc123xyz",
+		},
+		{
+			name:     "Valid old format",
+			revision: "main/12345678901234567890",
+			want:     "plan-main-1234567890",
+		},
+		{
+			name:     "Valid old format bucket format",
+			revision: "12345678901234567890",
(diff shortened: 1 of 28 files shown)


chanwit:
ping @mcortinas

api/planid/plan_id_test.go
```diff
@@ -0,0 +1,52 @@
...
+			name:     "Valid old format",
+			revision: "main/12345678901234567890",
+			want:     "plan-main-1234567890",
+		},
+		{
+			name:     "Valid old format bucket format",
+			revision: "12345678901234567890",
+			want:     "plan-1234567890",
+		},
+	}
```
Nalum, line 43: Should you also have some invalid cases?

Nalum (approved):
LGTM, just a comment on some extra tests

api/planid/plan_id_test.go
chanwit: The invalid case is the last one, where you will get the first ten characters as the suffix after `plan-`.

Nalum: Should it be named different then, as it says it's valid?

PR #757, opened 2023-07-11
feat: set approvePlan to be blank and force to be false

### LINKED ISSUE #751: Branch planner must set approvePlan to be `blank` and force to be `false` when cloning objects

yitsushi:
When cloning Terraform objects as Branch Planner Terraform objects, `approvePlan` has to be empty and `force` has to be `false` because they have no value, but they can mess up complext state-machine of the controller.

Closes #751

References:
* https://github.com/weaveworks/tf-controller/issues/751


chanwit (dismissed):
LGTM
Thank you @yitsushi :1st_place_medal:

PR #761, opened 2023-07-11
patch: use "flux-system" namespace instead of "default" for config

### LINKED ISSUE #760: The branch planner program would run inside `flux-system` by default, so configuration is assumed to be read from the `flux-system` instead of `default`.

yitsushi:
The branch planner program would run inside `flux-system` by default, so configuration is assumed to be read from the `flux-system` instead of `default`.

Closes #760

References:
* https://github.com/weaveworks/tf-controller/issues/760

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -27,7 +27,7 @@ import (
 //   secretName: bbp-token
 //   # List of Terraform resources
 //   resources: |-
-//     - namespace: default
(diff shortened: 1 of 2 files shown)


chanwit:
@yitsushi I forgot this one : `const DefaultConfigMapName = "flux-system/branch-planner"`

chanwit (approved):

PR #768, opened 2023-07-11
Improving ConfigMap Handling and Parsing Object Key, Removed Redundant Notation for the Controller and the Branch Planner

squaremo:
These are just some things I noticed looking at PRs.

 - configmap.Data is a map and can be nil, so check for that to avoid an NPE, and check the expected field is there so it can fail there and then rather than more mysteriously and later
 - be more precise when parsing an object key from a string
 - remove a bunch of unnecessary notation that made me have to pause and figure out why it was there. I still don't know, but it's not necessary, so to save the next person trying to figure it out, ...


chanwit (approved):

squaremo:
Thanks for giving the PR a better title @chanwit, that was not my best work :-)

PR #769, opened 2023-07-12
if force mode is true, we skip the plan annotation check

### LINKED ISSUE #745: Check the tfctl replan behaviour with auto approve and force = true
We need to document the behaviour of `tfctl replan` when `approvePlan="auto"` and `force=true`

chanwit:
Fixes #745

In response to user feedback, we are implementing a fix where the plan annotation check is bypassed if in force mode. This update is intended to mitigate the prevalent issue experienced by many users who were struggling with the Terraform controller bug. This bug was caused by a failure in parsing of the Flux revision format change from v0.41 to v2, and many users thought that changing `force: true` would forcefully apply the plan regardless and no error.

### DIFF
--- a/runner/server_load_tfplan.go
+++ b/runner/server_load_tfplan.go
@@ -48,7 +48,7 @@ func loadTFPlan(
 		return nil, err
 	}
 
-	if terraform.Spec.Force == true {
+	if terraform.Spec.Force {
 		// skip the annotation check
 		log.Info("force mode, skipping the plan's annotation check")
 	} else {
--- a/runner/server_load_tfplan.go
+++ b/runner/server_load_tfplan.go
@@ -1,0 +1,84 @@
+package runner
+
+import (
+	"context"
+	"fmt"
+	"github.com/go-logr/logr"
+	"path/filepath"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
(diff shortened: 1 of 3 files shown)


yiannistri (commented):
Can we add a test for this?

chanwit:
Thank you for bringing this up @yiannistri I'll add one

chanwit:
refactored the load plan function to make it testable.

runner/server_load_tfplan.go
```diff
@@ -0,0 +1,84 @@
...
+	tfplanSecretKey := types.NamespacedName{Namespace: req.Namespace, Name: "tfplan-" + terraform.WorkspaceName() + "-" + req.Name}
+	tfplanSecret := corev1.Secret{}
+	err := client.Get(ctx, tfplanSecretKey, &tfplanSecret)
+	if err != nil {
+		err = fmt.Errorf("error getting plan secret: %s", err)
+		log.Error(err, "unable to get secret")
+		return nil, err
+	}
+
+	if terraform.Spec.Force == true {
```
yiannistri, line 51:
```suggestion
	if terraform.Spec.Force {
```

yiannistri (approved):

yiannistri (approved):

PR #782, opened 2023-07-18
Document ADR for denying cross-ns refs by default

squaremo:
This PR introduces a decision record for changing the default treatment of cross-namespace references, which appear in a couple of places in the API.

This is a reversal of the (implicit) decision to respect cross-namespace references by default, so deserves to be recorded.

### DIFF
--- a/docs/adr/0002-deny-cross-ns-by-default.md
+++ b/docs/adr/0002-deny-cross-ns-by-default.md
@@ -1,0 +1,49 @@
+# 2. Deny cross-namespace refs by default
+
+* Status: [ **proposed** | rejected | accepted | deprecated ]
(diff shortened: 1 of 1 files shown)


squaremo:
@chanwit @Nalum @phoban01 @tomhuang12 As maintainers, would you check you're happy with this decision. With reference to https://github.com/weaveworks/tf-controller/blob/main/rfcs/DECISION_MAKING.md, I think lazy consensus is enough here; I'll merge in a couple of days, assuming at least one person approves the PR.

phoban01 (approved):

Nalum (approved):

chanwit (approved):
Thank you @squaremo 
LGTM

PR #788, opened 2023-07-20
add dependabot.yaml

### LINKED ISSUE #733: Enable GitHub dependent bot

chanwit:
Fixes #733

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,12 +1,18 @@
 version: 2
 
 updates:
-  - package-ecosystem: "github-actions"
+  - package-ecosystem: "gomod"
     directory: "/"
     labels: ["area/ci", "dependencies"]
     schedule:
-      # By default, this will be on a monday.
-      interval: "weekly"
+      interval: "daily"
+    # Project maintainers and the Wild Watermelon team
+    reviewers:
+      - chanwit
+      - Nalum
+      - phoban01
+      - tomhuang12
+      - "weaveworks/wild-watermelon"
     groups:
       # Group all updates together, so that they are all applied in a single PR.
       # Grouped updates are currently in beta and is subject to change.
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,16 @@
+version: 2
+
+updates:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,16 @@
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
```
enekofb, line 4: i believe that this would only for github-actions workflow based on https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem 

In case you want to enable it for go.mod dependencies you might need to change the [package-ecosystem](https://github.com/weaveworks/pipeline-controller/pull/190/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28)

.github/dependabot.yaml
```diff
@@ -0,0 +1,16 @@
...
+    directory: "/"
+    labels: ["area/ci", "dependencies"]
+    schedule:
+      # By default, this will be on a monday.
+      interval: "weekly"
```
enekofb, line 9: In the other repos we are setting it as daily https://github.com/weaveworks/pipeline-controller/pull/190/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R6 

I am not that daily is the right long-term choice (for example due to the amount of PRs raised) but adding it here in case you want to have the same settings as other repos.

.github/dependabot.yaml
```diff
@@ -0,0 +1,16 @@
...
+    groups:
```
enekofb, line 10: another useful option might be using [reviewers](https://github.com/weaveworks/pipeline-controller/pull/190/files#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R7) so PRs are assigned for reviewing reducing the risk or staying there without review for not being assigned? i guess that is up you to decide what works best

yitsushi (approved):

PR #801, opened 2023-07-25
feat: post error as comment to a pull request

### LINKED ISSUE #778: Handle the error, show the error message
**User Story:**

As a Branch planner user,
I'd like to have a clear error message displayed when an error occurs as comments of the current Pull Request,
so that I can quickly identify and fix any issues in my Terraform configuration.

**Acceptance Criteria:**
- [ ] Display a clear and comprehensive error message when an error occurs in the Terraform Planning object
- [ ] The error message should provide information on the cause of the error and possible steps for resolution

yitsushi:
When the informer gets a Terrafrom object with error state, it adds a new annotation to the object with the current revision. The informer is not able to see if it it's the same error or not.

Experimented with other available data like generation, the error message itself, store values as a condition, but as it's not a condition change, and we can assume the code will be the same if the revision is the same, I decided to go down that route and use the revision as an anchor point.

If we can't set the annotation, the system will NOT post the comment by design. I would rather be annoyed but having to debug tf-controller to see why the output plan/error is not available as a comment than getting 20-30 comments with the same error under my pull requests.

Closes #778

References:
* https://github.com/weaveworks/tf-controller/issues/778


yitsushi:
https://github.com/yitsushi/tf-controller-helloworld/pull/2
image

chanwit (approved):

PR #824, opened 2023-08-01
Prevent dependson crossnamespace reference

### LINKED ISSUE #694: Prevent dependsOn with another namespace when --no-cross-namespace-refs set, too
#677 added the flag `--no-cross-namespace-refs`, like the Flux controllers have, which makes the controller check `.spec.sourceRef` to make sure it's in the same namespace.
It should do the same for `.spec.dependsOn` refs, which also have a namespace field.

========

**User Story**

As a Terraform Controller user,
I'd like to have the option to prevent cross-namespace references when the flag `--no-cross-namespace-refs` is set,
so that I can ensure more secure interactions among resources and applications.

**Acceptance Criteria:**

- [ ] When `--no-cross-namespace-refs` is set, the controller should reject any `.spec.dependsOn` references that occur in a different namespace.
- [ ] When `--no-cross-namespace-refs` is set, the controller should continue to reject any `.spec.sourceRef` references that occur in a different namespace.
- [ ] If a resource does not meet the requirements of the `--no-cross-namespace-refs` flag, an appropriate error message should be provided.

luizbafilho:
closes #694 

- Makes sure the `dependsOn` functionality respects the no cross-namespace reference option.


controllers/tc000015_depends_on_cross_ns_denied_test.go
```diff
@@ -0,0 +1,126 @@
...
+				{Name: "gr-source", Namespace: "other-ns"},
+			},
+		},
+	}
+	It("should be created and attached successfully.")
+	g.Expect(k8sClient.Create(ctx, &helloWorldTF)).Should(Succeed())
+	t.Cleanup(func() { g.Expect(k8sClient.Delete(ctx, &helloWorldTF)).Should(Succeed()) })
+
+	It("should be access denied error.")
+	By("checking that the Ready's reason of the TF resource become `ArtifactFailed`.")
```
squaremo, line 108: `ArtifactFailed` does not seem like the right reason to give when you cannot access a dependsOn entry. Is there another suitable reason handy? You might need to create one.

controllers/tf_controller.go
```diff
@@ -232,6 +232,21 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	// check dependencies, if not being deleted
 	if len(terraform.Spec.DependsOn) > 0 && !isBeingDeleted(terraform) {
 		if err := r.checkDependencies(sourceObj, terraform); err != nil {
+			if acl.IsAccessDenied(err) {
+				traceLog.Info("The cross-namespace dependency was denied by reconciler.NoCrossNamespaceRefs")
+
+				terraform = infrav1.TerraformNotReady(terraform, sourceObj.GetArtifact().Revision, infrav1.ArtifactFailedReason, err.Error())
```
squaremo, line 238: Here too -- ArtifactFailed seems like the wrong reason (see comment in test case).

controllers/tf_controller.go
```diff
@@ -232,6 +232,21 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+				if err := r.patchStatus(ctx, req.NamespacedName, terraform.Status); err != nil {
+					log.Error(err, "unable to update status for source access denied")
```
squaremo, line 240:
```suggestion
					log.Error(err, "unable to update status for dependsOn access denied")
```

controllers/tf_controller.go
```diff
@@ -518,6 +533,12 @@ func (r *TerraformReconciler) SetupWithManager(mgr ctrl.Manager, maxConcurrentRe
 func (r *TerraformReconciler) checkDependencies(source sourcev1.Source, terraform infrav1.Terraform) error {
 	dependantFinalizer := infrav1.TFDependencyOfPrefix + terraform.GetName()
 	for _, d := range terraform.Spec.DependsOn {
+		if r.NoCrossNamespaceRefs && d.Namespace != terraform.GetNamespace() {
```
squaremo, line 536: I think you'll want this check to go after defaulting the namespace, since otherwise the second part will be true any time a dependsOn entry has an empty `.namespace` (which should always pass the check, since it denotes a resource in the same namespace).

squaremo: In fact, I think you're better off constructing `dName`, then defaulting the namespace, then checking if it violates NoCrossNamespaceRefs; that way it avoids mutating d, which makes me a little nervous.

squaremo (changes requested):
Nice and neat; thanks for including the test case.
There's one thing I think is a bug: see the comment on `checkDependencies`. Otherwise, take my comments as advisory.

controllers/tc000015_depends_on_cross_ns_denied_test.go
squaremo: I guess there's two ways to go here: say what failed (`ArtifactFailed`), or say why it failed (`AccessDenied`). "The object is not ready because [access was denied]" makes sense as a statement, so I'm happy with that.

controllers/tc000015_depends_on_cross_ns_denied_test.go
```diff
@@ -0,0 +1,126 @@
...
+			Name:      terraformName,
+			Namespace: "flux-system",
+		},
+		Spec: infrav1.TerraformSpec{
+			ApprovePlan: "auto",
+			Path:        "./terraform-hello-world-example",
+			SourceRef: infrav1.CrossNamespaceSourceReference{
+				Kind:      "GitRepository",
+				Name:      testRepo.Name,
+				Namespace: testRepo.Namespace,
```
squaremo, line 95: This should be (obviously) in the same namespace, so we don't confuse `AccessDenied` from the source ref being cross-ns with `AccessDenied` from the dependsOn being cross-ns.

squaremo: These are actually in the same namespace, so what I suggest is leaving the namespace out of the `SourceRef` value. Then if it breaks because of that, it'll look different.

squaremo (approved):
Little thing -- I think remove the namespace from the SourceRef in the test, so that it doesn't pass by coincidence. Then it's good to go :+1:

yitsushi (dismissed):

yitsushi (approved):

PR #840, opened 2023-08-03
Add flag --allow-cross-namespace-refs to tf-controller and branch-planner

squaremo:
With reference to [ADR 002](https://github.com/weaveworks/tf-controller/blob/main/docs/adr/0002-deny-cross-ns-by-default.md):

This adds the flag --allow-cross-namespace-refs, which defaults to `false` (do not allow cross namespace refs), and overrides the flag `--no-cross-namespace-refs` which is borrowed from Flux.

The rationale is that it's secure by default to not allow cross-namespace refs, and to make people explicitly allow them with the flag. It's early enough in the life of tf-controller to make this breaking change -- people using cross-namespace refs will just need to add the flag into their config.

Since people may already be using the flag `--no-cross-namespace-refs`, I've kept it. If `--allow-cross-namespace-refs` _is not_ supplied, and `--no-cross-namespace-refs` _is_ supplied, the latter will take effect. In practice this only matters if you pass `--no-cross-namespace-refs=false`, i.e., explicitly enable them via a double negative ("don't not enable them .."), since all other cases are the same as not passing either flag.

The implementation differs slightly between cmd/manager/ and cmd/branch-planner, just because of how these deal with flags. I've tried to follow a similar scheme, by passing a field `.NoCrossNamespaceRefs` to components. This also means a smaller change here, since that's how the option was conveyed when there was only `--no-cross-namespace-refs`.

Last thing: I did not write tests for this, on the basis that it's just flag parsing. I think it'd be reasonable to ask for some tests to make sure the right value ends up in `.NoCrossNamespaceRefs`, but I leave that decision to the reviewer.

### DIFF
--- a/cmd/branch-planner/flags.go
+++ b/cmd/branch-planner/flags.go
@@ -19,11 +19,12 @@ type applicationOptions struct {
 	allowedNamespaces []string
 
 	logOptions logger.Options
-	aclOptions acl.Options
 
 	runtimeNamespace   string
 	watchAllNamespaces bool
 	watchNamespace     string
+
+	noCrossNamespaceRefs bool
 }
 
 func parseFlags() *applicationOptions {
@@ -47,7 +48,13 @@ func parseFlags() *applicationOptions {
 		"Allowed namespaced. If it's empty, all namespaces are allowed for the planner. If it's not empty, only resources in the defined namespaces are allowed.")
 
 	opts.logOptions.BindFlags(flag.CommandLine)
-	opts.aclOptions.BindFlags(flag.CommandLine)
+
+	aclOptions := &acl.Options{}
+	aclOptions.BindFlags(flag.CommandLine)
+	// this flag exists so that the default is to _disallow_ cross-namespace refs. If supplied, it'll override `--no-cross-namespace-refs`; in other words, you can supply `--allow-cross-namespace-refs` with or without a value, and it will be observed.
+	var allowCrossNamespaceRefs bool
+	flag.BoolVar(&allowCrossNamespaceRefs, "allow-cross-namespace-refs", false,
+		"Enable following cross-namespace references. Overrides --no-cross-namespace-refs")
 
 	flag.Parse()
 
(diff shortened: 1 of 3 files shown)


cmd/branch-planner/flags.go
```diff
@@ -61,5 +68,12 @@ func parseFlags() *applicationOptions {
 		opts.watchNamespace = opts.runtimeNamespace
 	}
 
+	// as in ../manager/main.go, this is the case where --no-cross-namespace-refs can be different and not overridden.
```
yitsushi, line 71:
```
if
      (allow-cross-namespace-refs changed)
  and (no-cross-namespace-refs changed)
then
  allow-cross-namespace-refs = true
  no-cross-namespace-refs = true
  opts.noCrossNamespaceRefs := false
```

So `allow-cross-namespace-refs` overrides `no-cross-namespace-refs`. This comment implies `--no-cross-namespace-refs` can't be overridden.

Why not value check instead of checking if the value changed? usually I don't think I have issues with logical constructs, but I had to write notes to understand all cases properly as the "changed" is not the same as "if it's true". In this case technically yes, as both has `false` as default value, but still.

squaremo: > Why not value check instead of checking if the value changed?

```
	// --allow... \ --no... | true | false |  (false) |
	// ---------------------|------|-------|----|
	//     true             |  t   |   t   |  t |
	//     false            |  f   |   f   |  f |
	//     (false)          |  f   |   t*  |  f |
```

This is the table from cmd/manager/main.go, with the defaults in parentheses. The desired outcome for `--allow-cross-namespace-refs=false --no-cross-namespace-refs=false` is to disallow cross-namespace refs; but the desired outcome for `--no-cross-namespace-refs=false`, for which the flag values are the same as the previous case, is to allow cross-namespace refs.

It is annoyingly complicated. I think.

yitsushi: But what is the desired outcome if both set to true? As I understand it will disable `noCrossNamespaceRefs`, while the comment itself indicates (as I read) it should be `noCrossNamespaceRefs = true` because `noCrossNamespaceRefs` is explicitly set to `true` and it says "--no-cross-namespace-refs can be different and not overridden"

squaremo: If `--allow-cross-namespace-refs --no-cross-namespace-refs` the desired outcome is to allow cross-namespace refs (topmost, leftmost cell in the table). In this case, the conditional on L72 will evaluate to false and L75 will use the value from allowCrossNamespaceRefs, i.e., it will allow cross-namespace refs. So we agree on the desired outcome, and the code agrees too.

On the comment -- "this is the case where --no-cross-namespace-refs can be different and not overridden" -- this refers to the conditional checking that --no-cross-namespaces-refs is supplied but --allow-etc. is not (--> "not overridden"), giving a set of outcomes that differs to the other columns in the table. It's not a very useful comment. I'll change it.

yitsushi (dismissed):

squaremo:
(sorry @yitsushi there was a fixup I hadn't autosquashed, and a conflict with main)

Issue #462, opened 2022-11-27
Reconciler error on ipv6 cluster


dcshiman:
Hi, I am having a similar issue as [issues:365](https://github.com/weaveworks/tf-controller/issues/365)

Have tried setting the coreDNS with no luck, i am suspecting it's cos my cluster is on IPV6. Looking at the log, it looks like the tf controller is able to detect the status of runner pod, but can't figure out why its failing to run.

Here is the log form the controller, 
```
{"level":"info","ts":"2022-11-27T20:03:07.811Z","logger":"controller-runtime.metrics","msg":"Metrics server is starting to listen","addr":":8080"}
{"level":"info","ts":"2022-11-27T20:03:07.811Z","logger":"setup","msg":"Starting manager","version":"","sha":""}
{"level":"info","ts":"2022-11-27T20:03:07.812Z","msg":"Starting server","path":"/metrics","kind":"metrics","addr":"[::]:8080"}
{"level":"info","ts":"2022-11-27T20:03:07.812Z","msg":"Starting server","kind":"health probe","addr":"[::]:9440"}
I1127 20:03:07.913143       7 leaderelection.go:248] attempting to acquire leader lease flux-system/1953de50.contrib.fluxcd.io...
I1127 20:03:56.618185       7 leaderelection.go:258] successfully acquired lease flux-system/1953de50.contrib.fluxcd.io
{"level":"info","ts":"2022-11-27T20:03:56.618Z","msg":"Starting EventSource","controller":"terraform","controllerGroup":"infra.contrib.fluxcd.io","controllerKind":"Terraform","source":"kind source: *v1alpha1.Terraform"}
{"level":"info","ts":"2022-11-27T20:03:56.618Z","msg":"Starting EventSource","controller":"terraform","controllerGroup":"infra.contrib.fluxcd.io","controllerKind":"Terraform","source":"kind source: *v1beta2.GitRepository"}
{"level":"info","ts":"2022-11-27T20:03:56.618Z","msg":"Starting EventSource","controller":"terraform","controllerGroup":"infra.contrib.fluxcd.io","controllerKind":"Terraform","source":"kind source: *v1beta2.Bucket"}
{"level":"info","ts":"2022-11-27T20:03:56.618Z","msg":"Starting EventSource","controller":"terraform","controllerGroup":"infra.contrib.fluxcd.io","controllerKind":"Terraform","source":"kind source: *v1beta2.OCIRepository"}
[...]

Issue #365, opened 2022-09-27
Stuck in "Reconciliation in progress" state


dcfranca:
Hey everyone,

I'm just starting using Flux + TF controller, so please let me know if I'm missing some basic step
I have it running apparently properly for Kustomization, however, the `tf-controller` keeps hanging in `Reconciliation in progress`

I have created the GitRepo resource, pointing to my repository and correct path
The resource was created successfully and it is in `Ready` state

but the Terraform resource is not, and stays forever in `Reconciliation in progress` 

Here is my manifests: (the Terraform state is an existing one)
```
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: myrepo
  namespace: flux-system
spec:
  interval: 1m
  url: https://github.com/MyOrg/MyRepo.git
  ref:
    branch: master
  secretRef:
    name: flux-system
---
apiVersion: infra.contrib.fluxcd.io/v1alpha1
kind: Terraform
metadata:
  name: saas-github
  namespace: flux-system
spec:
  interval: 1m
  approvePlan: "disable"
  backendConfig:
    customConfiguration: |
      backend "s3" {
        bucket                      = "my-state-bucket"
        key                         = "my-bucket-key"
        region                      = "eu-west-1"
        dynamodb_table              = "lock-table"
        role_arn                    = "arn:aws:iam::XXXXXXX:role/role"
        encrypt                     = true
      }
  path: ./terraform/path
  sourceRef:
    kind: GitRepository
    name: myrepo
    namespace: flux-system
```

On the `flux-system` namespace I see those pods in running state

* helm-controller-xxxx
* kustomize-controller-xxxx
* notification-controller-xxxx
* saas-github-tf-runner
* source-controller-xxxxx
* tf-controller-xxxxx

And from the logs of tf-controller:
```
{"level":"info","ts":"2022-09-27T11:46:43.650Z","logger":"controller.terraform","msg":"getting source","reconciler group":"infra.contrib.fluxcd.io","reconciler kind":"Terraform","name":"saas-github","namespace":"flux-system"}
{"level":"info","ts":"2022-09-27T11:46:43.656Z","logger":"controller.terraform","msg":"trigger namespace tls secret generation","reconciler group":"infra.contrib.fluxcd.io","reconciler kind":"Terraform","name":"saas-github","namespace":"flux-system"}
[...]

PR #886, opened 2023-08-20
Allow DNS resolution of the runner pod for all k8s setup

syalioune:
Draft PR related to https://github.com/weaveworks/tf-controller/discussions/843#discussion-5493659 which content I'll replicate here.

First of all, thanks for the great work on this controller.
We've tried deploying recently the `tf-controller` in a standard `GKE clusters` and hit a wall because of DNS resolution issues. The controller is not able to resolve DNS names like `10-40-129-81.podinfo.pod.cluster.local`.

This issue is best described here https://github.com/weaveworks/tf-controller/issues/365#issuecomment-1262600671 and here https://github.com/weaveworks/tf-controller/issues/462#issuecomment-1329591240. GKE uses [Cloud DNS](https://cloud.google.com/dns?hl=fr) which does not provide IP based DNS name resolution like `CoreDNS`.

It's still very rough but revolves around :

- Setup a `tf-runner` service in each allowed namespaces
- Add `hostname=terraform_crd_name, subdomain=tf-runner` fields to the runner pod so that an `A record : terraform_crd_name.tf-runner.namespace.svc.cluster_domain` is automatically created
- Add a `SAN` into the runner generated certificate

Preliminary tests shows that it works.

Before going further, I'm looking for community/maintainer feedbacks :smile:

- Is this proposal in line with the project direction ?
- Should that method replace entirely pod's ip based resolution or be an alternative with an opt-in flag/option ?
- Should the controller be responsible for : checking the service existence in the allowed namespace ? creating the service if it does not exist ?

Cheers


charts/tf-controller/templates/runner-discovery-service.yaml
```diff
@@ -0,0 +1,16 @@
...
+  namespace: {{ . }}
+spec:
+  clusterIP: None
+  ports:
+  - name: grpc
+    port: 30000
+  selector:
+    app.kubernetes.io/created-by: tf-controller
+		app.kubernetes.io/name: tf-runner
+{{- end }}
```
chanwit, line 16: Hi @syalioune 

We want to rely only on Pod hostnames without introducing the Service layer.
Would it be possible to do so with your hacks?

chanwit (changes requested):

controllers/tf_controller_runner.go
```diff
@@ -233,6 +234,8 @@ func (r *TerraformReconciler) runnerPodSpec(terraform infrav1.Terraform, tlsSecr
 	}
 
 	return v1.PodSpec{
+		Hostname:                      terraform.Name,
+		Subdomain:                     "tf-runner",
```
chanwit, line 238: Maybe we introduce `--use-pod-subdomain-resolution` to guard here.

mtls/rotator.go
```diff
@@ -643,8 +643,11 @@ func (cr *CertRotator) generateNamespaceTLS(namespace string) (*corev1.Secret, e
 	artifactCache := cr.artifactCaches[n-1]
 	caArtifacts := artifactCache.ca
 
-	hostname := fmt.Sprintf("*.%s.pod.%s", namespace, cr.ClusterDomain)
-	cert, key, err := cr.createCertPEM(caArtifacts, hostname, time.Now().Add(-1*time.Hour), caArtifacts.validUntil)
+	hostnames := []string{
+		fmt.Sprintf("*.%s.pod.%s", namespace, cr.ClusterDomain),
+		fmt.Sprintf("*.tf-runner.%s.svc.%s", namespace, cr.ClusterDomain),
```
chanwit, line 648: Maybe we introduce `--use-pod-subdomain-resolution` to also guard here.

charts/tf-controller/templates/runner-discovery-service.yaml
```diff
@@ -0,0 +1,16 @@
...
+  name: tf-runner
+  namespace: {{ . }}
+spec:
+  clusterIP: None
+  ports:
+  - name: grpc
+    port: 30000
+  selector:
+    app.kubernetes.io/created-by: tf-controller
+		app.kubernetes.io/name: tf-runner
```
chanwit, line 15:
```suggestion
  selector:
    app.kubernetes.io/created-by: tf-controller
    app.kubernetes.io/name: tf-runner
```

chanwit: Please introduce the new chart value called `usePodSubdomainResolution` to
1. add `--use-pod-subdomain-resolution` flag to the controller (which requires you to create that flag for it too)
2. guard generation of `Service` objects here.

Please note that we need `tf-runner` Service by default for `flux-system` namespace too.

chanwit (changes requested):

chanwit:
Thank you for your contributions @syalioune 

Maybe we want to go with the subdomain resolution as you suggested.
And we also need a new flag (see above comments) for the controller to make sure this new feature does not affect the current users.

charts/tf-controller/templates/runner-discovery-service.yaml
syalioune: A service is mandatory in my proposal to have a consistent DNS entry generated regardless of the k8s cluster setup.
A service with pod hostnames/subdomains is the best compromise since it requires only one service per namespace instead of a service per runner pod.

controllers/tf_controller_runner.go
syalioune: Yeah, will do !

mtls/rotator.go
syalioune: Yeah, will do !

controllers/tf_controller_runner.go
syalioune: Those two fields are now guarded by `--use-pod-subdomain-resolution`

mtls/rotator.go
syalioune: Done

charts/tf-controller/templates/runner-discovery-service.yaml
syalioune: Modifications performed :smile:

chanwit (dismissed):
LGTM
Thank you so much @syalioune :1st_place_medal:

cmd/manager/main.go
```diff
@@ -114,6 +115,7 @@ func main() {
 	flag.IntVar(&runnerGRPCMaxMessageSize, "runner-grpc-max-message-size", 4, "The maximum message size for gRPC connections in MiB.")
 	flag.BoolVar(&allowBreakTheGlass, "allow-break-the-glass", false, "Allow break the glass mode.")
 	flag.StringVar(&clusterDomain, "cluster-domain", "cluster.local", "The cluster domain used by the cluster.")
+	flag.BoolVar(&usePodSubdomainResolution, "use-pod-subdomain-resolution", false, "Allow to use pod hostname/ssubdomain DNS resolution instead of IP based")
```
souleb, line 118:
```suggestion
	flag.BoolVar(&usePodSubdomainResolution, "use-pod-subdomain-resolution", false, "Allow to use pod hostname/subdomain DNS resolution instead of IP based")
```

charts/tf-controller/templates/runner-discovery-service.yaml
```diff
@@ -0,0 +1,18 @@
...
+spec:
+  clusterIP: None
+  ports:
+  - name: grpc
+    port: 30000
+  selector:
+    app.kubernetes.io/created-by: tf-controller
+    app.kubernetes.io/name: tf-runner
+{{- end }}
+{{- end }}
```
souleb, line 18: the yaml file needs a new line I believe.

chanwit (approved):

charts/tf-controller/values.yaml
```diff
@@ -103,6 +103,9 @@ allowBreakTheGlass: false
 # -- Argument for `--cluster-domain` (Controller).
 #  ClusterDomain indicates the cluster domain, defaults to cluster.local.
 clusterDomain: cluster.local
+# -- Argument for `--use-pod-subdomain-resolution` (Controller).
+#  UsePodSubdomainResolution allow pod hostname/subdomain DNS resolution for the pod runner instead of IP based DNS resolution.
+usePodSubdomainResolution: false
```
chanwit, line 108:
```suggestion
usePodSubdomainResolution: false
```

chanwit (changes requested):

chanwit (approved):

PR #903, opened 2023-08-28
Replacing RunnerPodSpec with corev1.PodSpec

### LINKED ISSUE #786: Ability to set `priorityClass` on tf-runner pod
Hi,
we have the situation that we want to run around 100 Microservice TF stacks with ephemeral runners. Those stacks are not huge but still require CPU and Memory at `terraform init` and `terraform plan`.

As we have no influence on the schedule for these TF stacks, we would like to have a k8s quota on eg. number of pods in the `flux-system` namespace.
This would be doable but inflicts the risk that we put crucial pods back in the queue and render the Flux system unusable.

My thoughts went to `priorityClass` as a distinctor but I can not set the separate class as I lack access to the pod template. Hence my question: could you please somehow provide an interface to influence the pods deeper? Similar to the request on Resources and Limits (#526).

Greetings and thanks for this piece of software!

- https://kubernetes.io/docs/concepts/policy/resource-quotas/
- https://github.com/weaveworks/tf-controller/blob/19027da589d05e78165972807cbfa2553782a90a/controllers/tf_controller_runner.go#L40

========

**User Story:**

As a Kubernetes cluster administrator,
I'd like to be able to set a `priorityClass` on the tf-runner pod,
so that I can manage the scheduling and resource quotas of Terraform workloads efficiently.

**Acceptance Criteria:**

- [ ] Users can set a `priorityClass` for the tf-runner pod through the Terraform Controller configuration
- [ ] The `priorityClass` setting is properly applied to the tf-runner pod when scheduled
- [ ] The Terraform workloads adhere to the scheduling rules and resource quotas enforced by Kubernetes

luizbafilho:
fixes #786 

To allow full customization of the runner pod, this PR replaces the custom `RunnerPodSpec` type with `corev1.PodSpec`, and uses the `mergo` package to merge the default settings with the user's ones.

### DIFF
--- a/api/v1alpha2/reference_types.go
+++ b/api/v1alpha2/reference_types.go
@@ -143,7 +143,7 @@ type RunnerPodTemplate struct {
 	Metadata RunnerPodMetadata `json:"metadata,omitempty"`
 
 	// +optional
-	Spec RunnerPodSpec `json:"spec,omitempty"`
+	Spec corev1.PodSpec `json:"spec,omitempty"`
 }
 
 type RunnerPodMetadata struct {
--- a/config/crd/bases/infra.contrib.fluxcd.io_terraforms.yaml
+++ b/config/crd/bases/infra.contrib.fluxcd.io_terraforms.yaml
@@ -5287,9 +5287,17 @@ spec:
                         type: object
                     type: object
                   spec:
+                    description: PodSpec is a description of a pod.
                     properties:
+                      activeDeadlineSeconds:
+                        description: Optional duration in seconds the pod may be active
+                          on the node relative to StartTime before the system will
+                          actively try to mark it failed and kill associated containers.
+                          Value must be a positive integer.
+                        format: int64
+                        type: integer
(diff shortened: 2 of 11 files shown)


squaremo:
I understand the motivation, and I can see the problem-- `PodSpec` has required fields, and as soon as you are supplying a `PodSpec` value you will have to provide all of those. What you need is `PodSpec` but with everything marked optional, and I can think of some ways of getting that:

 1. make a copy of the PodSpec type locally and change all the annotations
 2. patch the generated CRD schema so that the fields are all optional
 3. use a freeform-JSON type for the field

All of these have obvious problems. Patching the schema might have the least serious problems, since schema changes can be caught during build.

squaremo:
4. If it's not a requirement to be able to patch anything in the pod spec, then just including the fields known to be required would be fine.

yiannistri:
Closing in favour of https://github.com/weaveworks/tf-controller/pull/910

Issue #865, opened 2023-08-14
[Umbrella issue] Investigate, Fix and turn flaky tests on


yitsushi:
Feel free to create new sub-issues.

  - [x] #908
  - [x] #920
  - [x] #904
  - [x] #905
  - [x] #906
  - [x] #907
  - [x] #926
  - [x] #932
  - [x] #945
  - [x] #946
  - [x] #947
  - [x] #948
  - [x] #949
  - [x] #950
  - [x] #951
  - [x] #952
  - [x] #954

LappleApple:
Relates to https://github.com/weaveworks/tf-controller/pull/863

yiannistri:
We suspect that this is caused by the cluster state not being reset after each test run.

yitsushi:
Note for testing:

1. Run the test once. If it fails, you see what's the error.
2. Run the test with `-count`: `go test ./controllers -parallel 1 -count=3 -test.run Test_XXXXX` [^1]
3. If everything is happy, run with all the tests in the same "group" (group of 10s) together with `-count` like `go test ./controllers -parallel 1 -count=3 -test.run Test_00004` for all the `04x` tests

[^1]: to use `go test` you need some env vars, check makefile, but these should be enough: `export INSECURE_LOCAL_RUNNER=1; export DISABLE_K8S_LOGS=1; export DISABLE_TF_LOGS=1; export DISABLE_TF_K8S_BACKEND=1; export DISABLE_WEBHOOK_TLS_VERIFY=1; export KUBEBUILDER_ASSETS="$(pwd)/testbin/k8s/1.28.0-linux-amd64"`

PR #938, opened 2023-09-08
experiment: break up test jobs

yitsushi:
Benefits:
- Easier to re-run failed jobs, it does not re-run all of them
- Each group has it's own envtest
- Because it's parallel, it's faster (see comment)

related to #865


yitsushi:
It's even faster. Comparison:

Previous:
image
image


This one:
image

PR #959, opened 2023-09-12
Screen cross-ns refs in .spec.cliConfigSecretRef

### LINKED ISSUE #957: Apply cross-namespace reference ban to `.spec.cliConfigSecretRef`
Another cross-namespace ref! In https://github.com/weaveworks/tf-controller/blob/v0.16.0-rc.2/api/v1alpha2/terraform_types.go#L180, and handled in https://github.com/weaveworks/tf-controller/blob/v0.16.0-rc.2/controllers/tf_controller_backend.go#L151. This hsould be subject to the NoCrossNamespaceRefs option.

squaremo:
This is another place that a cross-namespace ref can be supplied. This I learned: corev1.SecretReference has a namespace field.
Fixes #957.

### DIFF
--- a/controllers/tc000015_cliconfigsecret_cross_ns_denied_test.go
+++ b/controllers/tc000015_cliconfigsecret_cross_ns_denied_test.go
@@ -1,0 +1,126 @@
+package controllers
+
+import (
+	"context"
+	"testing"
+	"time"
+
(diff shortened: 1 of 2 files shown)


yiannistri (approved):
:+1:

PR #964, opened 2023-09-13
Adds page on upgrading TF-Controller

LappleApple:
Page added by user request. Steps reviewed and updated by @madAndroid.

### DIFF
--- a/docs/how_to/upgrade-tf-controller.md
+++ b/docs/how_to/upgrade-tf-controller.md
@@ -7,8 +7,8 @@ Please follow these steps to upgrade TF-Controller:
 3. Suspend related Kustomization objects to minimize impact on live systems.
 4. Upgrade Flux first, following [the Flux documentation](https://fluxcd.io/flux/installation/upgrade/).
 5. Back up Terraform tfstates to avoid losing data. If you're using the default backend with secrets in Kubernetes, use your backup toolset (i.e., Velero) to back up the state data.
-6. To prevent unintentional resource deletions, set the 'spec.destroyResourcesOnDeletion' flag to `False` for critical or production systems.
-7. If the Flux upgrade goes well, proceed to upgrade the tf-controller via its image tag. Adjust the values in the HelmRelease to match the new version to which you are upgrading.
+6. To prevent unintentional resource deletions, set the `spec.destroyResourcesOnDeletion` flag to `False` for critical or production systems.
+7. If the Flux upgrade goes well, proceed to upgrade the TF-controller via its image tag. Adjust the values in the HelmRelease to match the new version to which you are upgrading.
 8. Check your system logs to identify any potential issues.
 9. Push the changes you made.
 10. Ensure no changes are planned for deletion. TF-Controller has a flag to help prevent the deletion: `spec.destroyResourcesOnDeletion`. This is set to `false` by default.
--- a/docs/how_to/upgrade-tf-controller.md
+++ b/docs/how_to/upgrade-tf-controller.md
@@ -7,15 +7,17 @@ Please follow these steps to upgrade TF-Controller:
 3. To make sure you don't get new state changes, suspend Terraform resources (`tfctl suspend --all`) to minimize the impact on live systems.
 4. Back up Terraform tfstates to avoid losing data. If you're using the default backend with secrets in Kubernetes, use your backup toolset (i.e., Velero) to back up the state data.
 5. Upgrade Flux first, following [the Flux documentation](https://fluxcd.io/flux/installation/upgrade/).
-6. To prevent unintentional resource deletions, set the `spec.destroyResourcesOnDeletion` flag to `false` for critical or production systems (the default value is `false`)
-7. If the Flux upgrade goes well, proceed to upgrade the TF-controller via its image tag. Adjust the values in the HelmRelease to match the new version to which you are upgrading.
-8. Check the pod logs for the TF-Controller deployment and any runner logs in order to identify potential issues. If you check the `warnings` in the logs, you can also identify any required API changes. For example:
+6. Disable [auto-approval](https://weaveworks.github.io/tf-controller/use_tf_controller/to_provision_resources_and_auto_approve/) by either removing the approvePlan value or setting it to "".
+7. To prevent unintentional resource deletions, set the `spec.destroyResourcesOnDeletion` flag to `false` for critical or production systems (the default value is `false`)
+8. If the Flux upgrade goes well, proceed to upgrade the TF-controller via its image tag. Adjust the values in the HelmRelease to match the new version to which you are upgrading.
+9. Check the pod logs for the TF-Controller deployment and any runner logs in order to identify potential issues. If you check the `warnings` in the logs, you can also identify any required API changes. For example:
 ` v1alpha1 Terraform is deprecated, upgrade to v1alpha2`.
-9. Push the changes you made.
-10. Resume your Terraform resources—either one-by-one for critical resources, or all of them with `tfctl resume --all`
-11. Ensure no changes are planned for deletion. If you changed the value in step 6 from `spec.destroyResourcesOnDeletion` to `false`, resources will not be automatically removed.
-12. Revert back to auto-approval mode after ensuring stability.
-13. Resume any suspended Kustomization objects to restore GitOps automation.
+10. Push the changes you made.
+11. Resume your Terraform resources—either one-by-one for critical resources, or all of them with `tfctl resume --all`
+12. Ensure no changes are planned for deletion. If you changed the value in step 6 from `spec.destroyResourcesOnDeletion` to `false`, resources will not be automatically removed.
+13. Revert back to auto-approval mode after ensuring stability.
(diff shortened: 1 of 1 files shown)


docs/how_to/upgrade-tf-controller.md
```diff
@@ -0,0 +1,16 @@
+# Upgrading TF-Controller
+
+Please follow these steps to upgrade TF-Controller:
+
+1. Read the latest release changelogs.
+2. Check your API versions.
+3. Suspend related Kustomization objects to minimize impact on live systems.
+4. Upgrade Flux first, following [the Flux documentation](https://fluxcd.io/flux/installation/upgrade/).
+5. Back up Terraform tfstates to avoid losing data. If you're using the default backend with secrets in Kubernetes, use your backup toolset (i.e., Velero) to back up the state data.
```
yitsushi, line 9: I would swap these two. If Flux was updated, but the new Flux version is not supported by the currently TF Controller, the controller can behave weirdly and safer to save states before it happens.

For example upgrading from v0.14 to v0.15. That requires a new Flux version. Either way (upgrade flux or tf controller first) it will throw errors if the controller reconciliation loop starts.

4. Suspend terraform resources. (`tfctl suspend --all`)   << this needs to be done before backup to make sure we don't get new state changes.
5. Back up your data.
6. Upgrade Flux
...
N. (probably after current `8.` or `9.`) Resume terraform resources. Either one-by-one for critical resources, or YOLO `tfctl resume --all`

yitsushi: What do you think @chanwit ?

madAndroid: @yitsushi - Thanks for outlining the additional steps required - I'm going to look at testing these once the list of steps has been confirmed, will keep watching here until they're finalised.

madAndroid: Also worth noting here: it might be prudent to include any upgrade steps, specific to a particular release, in the release notes for that release? for example, you've noted:

> For example upgrading from v0.14 to v0.15. That requires a new Flux version

These specifics are best noted in the release notes, IMO - I'm not sure if that's currently the case ... this would answer a question a user had: "How do I know when to update `apiVersion` for TF CRs when upgrading" .. we could potentially point them to specific release notes _per release_ in this instance - I hope that makes sense? :)

yitsushi: The example I provided was to support my additions, to show how it can behave weirdly. It shouldn't cause much issues with states, but we definitely want to make backup before upgrading Flux.

LappleApple: @chanwit Would you reverse steps 4 and 5 as @yitsushi suggests? LMK, and then I'll wrap this up.

LappleApple: @chanwit Yes, I've already been in touch with @madAndroid about testing them

LappleApple: @yitsushi @chanwit Updated based on your comments -- thanks, both. @madAndroid, want to give a test run?

madAndroid: @chanwit - thanks for the clarification :) is it worth having these details present in the release notes? it would ideally be easily reachable by clients, so that we can refer them to that location when they have questions around apiVersion upgrades

madAndroid: Apologies, I do see that the CHANGELOG does include details around breaking changes and the like, so would this be the correct location to point customers? => https://github.com/weaveworks/tf-controller/blob/main/CHANGELOG.md

madAndroid: Re: testing, I will make sure the steps are followed ASAP - I have a test cluster I can use for this

yiannistri: @madAndroid did you have a chance to test this?

docs/how_to/upgrade-tf-controller.md
```diff
@@ -0,0 +1,22 @@
...
+4. Back up Terraform tfstates to avoid losing data. If you're using the default backend with secrets in Kubernetes, use your backup toolset (i.e., Velero) to back up the state data.
+5. Upgrade Flux first, following [the Flux documentation](https://fluxcd.io/flux/installation/upgrade/).
+6. To prevent unintentional resource deletions, set the `spec.destroyResourcesOnDeletion` flag to `False` for critical or production systems.
+7. If the Flux upgrade goes well, proceed to upgrade the TF-controller via its image tag. Adjust the values in the HelmRelease to match the new version to which you are upgrading.
+8. Check your system logs to identify any potential issues.
+9. Push the changes you made.
+10. Resume your Terraform resources—either one-by-one for critical resources, or all of them with `tfctl resume --all`
+11. Ensure no changes are planned for deletion. TF-Controller has a flag to help prevent the deletion: `spec.destroyResourcesOnDeletion`. This is set to `false` by default.
+12. Revert back to auto-approval mode after ensuring stability.
```
madAndroid, line 16: Looks like we're missing a step for disabling `approvePlan` further up the chain? between 5 and 6, there should be a step to either remove the `approvePlan` value, or set it to: `""` - we're stating in this step that we need to revert back, but there hasn't been a step to disable auto-approval (perhaps a link to the [docs](https://weaveworks.github.io/tf-controller/use_tf_controller/to_provision_resources_and_auto_approve/)  would be useful?

docs/how_to/upgrade-tf-controller.md
```diff
@@ -0,0 +1,22 @@
...
+13. Resume any suspended Kustomization objects to restore GitOps automation.
```
madAndroid, line 17: An additional step would be to restore `spec.destroyResourcesOnDeletion`, if this has been disabled for any resources in critical or production systems (ie, reverting the change made in step 6)

madAndroid (changes requested):
A few minor tweaks suggested, to make the defined steps more logical

docs/how_to/upgrade-tf-controller.md
```diff
@@ -0,0 +1,22 @@
...
+Please follow these steps to upgrade TF-Controller:
+
+1. Read the latest release changelogs.
+2. Check your API versions.
+3. To make sure you don't get new state changes, suspend Terraform resources (`tfctl suspend --all`) to minimize the impact on live systems.
+4. Back up Terraform tfstates to avoid losing data. If you're using the default backend with secrets in Kubernetes, use your backup toolset (i.e., Velero) to back up the state data.
+5. Upgrade Flux first, following [the Flux documentation](https://fluxcd.io/flux/installation/upgrade/).
+6. To prevent unintentional resource deletions, set the `spec.destroyResourcesOnDeletion` flag to `False` for critical or production systems.
+7. If the Flux upgrade goes well, proceed to upgrade the TF-controller via its image tag. Adjust the values in the HelmRelease to match the new version to which you are upgrading.
+8. Check your system logs to identify any potential issues.
```
madAndroid, line 12:
```suggestion
8. Check the pod logs for the tf-controller deployment, as well as any runner logs, to identify any potential issues.
```

docs/how_to/upgrade-tf-controller.md
```diff
@@ -0,0 +1,22 @@
...
+9. Push the changes you made.
+10. Resume your Terraform resources—either one-by-one for critical resources, or all of them with `tfctl resume --all`
+11. Ensure no changes are planned for deletion. TF-Controller has a flag to help prevent the deletion: `spec.destroyResourcesOnDeletion`. This is set to `false` by default.
```
madAndroid, line 15:
```suggestion
11. Ensure no changes are planned for deletion - if you changed the value in step 6, for `spec.destroyResourcesOnDeletion` to `false`, resources will not be automatically removed.
```

docs/how_to/upgrade-tf-controller.md
```diff
@@ -0,0 +1,22 @@
+# Upgrading TF-Controller
+
+Please follow these steps to upgrade TF-Controller:
+
+1. Read the latest release changelogs.
+2. Check your API versions.
+3. To make sure you don't get new state changes, suspend Terraform resources (`tfctl suspend --all`) to minimize the impact on live systems.
+4. Back up Terraform tfstates to avoid losing data. If you're using the default backend with secrets in Kubernetes, use your backup toolset (i.e., Velero) to back up the state data.
+5. Upgrade Flux first, following [the Flux documentation](https://fluxcd.io/flux/installation/upgrade/).
+6. To prevent unintentional resource deletions, set the `spec.destroyResourcesOnDeletion` flag to `False` for critical or production systems.
```
madAndroid, line 10:
```suggestion
6. To prevent unintentional resource deletions, set the `spec.destroyResourcesOnDeletion` flag to `false` for critical or production systems (the default value is `false`)
```

yitsushi (approved):

LappleApple:
Applied changes requested by @madAndroid

madAndroid:
> I should be able to have the steps validated by midday tomorrow at the latest.

Testing all complete, and steps have been validated, this PR is good to go now 👍

PR #966, opened 2023-09-13
Explain `allowCrossNamespaceRefs` in chart README and docs/howto/

### LINKED ISSUE #956: Document --allow-cross-namespace-refs
- [x] explain how cross-namespace refs are handled, and the flags --allow-cross-namespace-refs and --no-cross-namespace-refs
 - [x] document the value in the Helm chart README

squaremo:
Explains that cross-namespace refs are not allowed by default, and how to allow them.

I wanted to mention when cross-namespace refs started being disallowed by default, so I guessed at when it would appear in a release -- version >0.16, since it would indicate a minor version bump at least.

Fixes #956.

### DIFF
--- a/charts/tf-controller/README.md
+++ b/charts/tf-controller/README.md
@@ -18,6 +18,10 @@ helm upgrade -i tf-controller tf-controller/tf-controller \
     --namespace flux-system
 ```
 
+### Using cross-namespace references
+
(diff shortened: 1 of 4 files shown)


yiannistri (dismissed):

squaremo:
> I wanted to mention when cross-namespace refs started being disallowed by default, so I guessed at when it would appear in a release -- version >0.16, since it would indicate a minor version bump at least.

I think I have this wrong -- the breaking change is already merged, and the next minor release is going to be 0.16.0 (since we're on RCs now). So I should make the cited version `=0.16.0`.

yiannistri (dismissed):

yiannistri (approved):

PR #1001, opened 2023-09-19
Deleting the Terraform object before deleting its source

### LINKED ISSUE #898: Branch Planner's source being deleted before the planning object
There appears to be a bug in the branch planner. When a PR is merged, the GitRepository is deleted before the associated terraform object. This sequence leads to the terraform object deletion being stuck due to the main issue.

chanwit:
Fixes #898 

This PR contains two commits.

1. The first commit fixed the Terraform deletion logic
of the Branch Planner by making sure that the
Terraform object is deleted before the source.

It also changed the plan object creation logic by
making sure that
- the plan object won't delete the live resources.
- the plan object won't create any output secrets.

2. There was a bug inside the createLabels function
that modified the labels of the original objects
instead of creating a new map. The second commit fixed it.


yitsushi (approved):
:shipit:

PR #1010, opened 2023-09-20
propose adr 0003 for workspace blob caching

### LINKED ISSUE #972: Write an ADR for the workspace blob caching
From this discussion:

https://github.com/weaveworks/tf-controller/discussions/911

chanwit:
Fixes #972

### DIFF
--- a/docs/adr/0003-workspace-blob-caching.md
+++ b/docs/adr/0003-workspace-blob-caching.md
@@ -25,7 +25,8 @@ and ease of access.
      to compress the Workspace file system into a tar.gz format, which is then retrieved
      as a byte array.
    * The caching mechanism will be executed right before the Terraform Initialization step, ensuring that the latest and most relevant data is used.
-   * Each Workspace Blob will be cached on the TF-Controller's local disk, following the naming convention `$namespace-$name.tar.gz`.
+   * Each Workspace Blob will be cached on the TF-Controller's local disk, using the UUID of the Terraform object as the filename,`${uuid}.tar.gz`.
+   * To prevent unauthorized access to the cache entries, and cache collisions, the cache file will be deleted after the finalization process is complete.
 2. **Persistence** 
    * The persistence mechanism used by the Source Controller will be adopted for the TF-Controller's persistence volume.
 3. **BLOB Encryption**
--- a/docs/adr/0003-workspace-blob-caching.md
+++ b/docs/adr/0003-workspace-blob-caching.md
@@ -26,17 +26,16 @@ A caching mechanism for these BLOBs is essential to fixing the single TF object 
 
 1. **BLOB Creation and Storage**
    * A gRPC function named `CreateWorkspaceBlob` will be invoked by the TF-Controller 
-     to compress the Workspace file system into a tar.gz format, which is then retrieved
-     as a byte array.
+     to tell tf-runner to compress the Workspace file system into a tar.gz BLOB, which is then retrieved back to the controller.
    * The caching mechanism will be executed right before the Terraform Initialization step, ensuring that the latest and most relevant data is used.
    * Each Workspace Blob will be cached on the TF-Controller's local disk, using the UUID of the Terraform object as the filename,`${uuid}.tar.gz`.
-   * To prevent unauthorized access to the cache entries, and cache collisions, the cache file will be deleted after the finalization process is complete.
+   * To reduce the risk of unauthorized access to the cache entries, and cache collisions, the cache file will be deleted after the finalization process is complete.
 2. **Persistence** 
-   * The persistence mechanism used by the Source Controller will be adopted for the TF-Controller's persistence volume.
+   * [The persistence mechanism used by the Source Controller](https://fluxcd.io/flux/installation/configuration/vertical-scaling/#persistent-storage-for-flux-internal-artifacts) will be adopted for the TF-Controller's persistence volume.
 3. **BLOB Encryption**
    * The encryption and decryption of the BLOBs will be tasked to the runner, with the controller solely responsible for storing encrypted BLOBs.
    * Each namespace will require a service account, preferably named "tf-runner".
-   * The token of this service account, which is natively supported by Kubernetes, will serve as the most appropriate encryption key.
+   * The token of this service account, which is natively supported by Kubernetes, will serve as the most appropriate encryption key because it's stored in a Secret, access to which can be controlled by RBAC. Storing it in a Secret also allows the key to be rotated.
 4. **Security Measures (Based on STRIDE Analysis)**
    * **Spoofing:** Implement Kubernetes RBAC for access restrictions and use mutual authentication for gRPC communications.
    * **Tampering:** Use checksums for integrity verification and 0600 permissions to write-protect local disk storage.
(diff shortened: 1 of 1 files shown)


docs/adr/0003-workspace-blob-caching.md
```diff
@@ -0,0 +1,54 @@
...
+and ease of access.
+
+## Decision
+
+1. **BLOB Creation and Storage**
+   * A gRPC function named `CreateWorkspaceBlob` will be invoked by the TF-Controller 
+     to compress the Workspace file system into a tar.gz format, which is then retrieved
+     as a byte array.
+   * The caching mechanism will be executed right before the Terraform Initialization step, ensuring that the latest and most relevant data is used.
+   * Each Workspace Blob will be cached on the TF-Controller's local disk, following the naming convention `$namespace-$name.tar.gz`.
```
yitsushi, line 28: By default it seems reasonable, but I would like to note, it can still cause scenarios when we still hit the "unable to delete" issue. If the controller container was restarted while the source was deleted (for example the controller crashed or rescheduled on a different node), it will not able to retrieve blob cache or blob artifact data from source.

yitsushi: Another note: in the `finalizer` we have to delete cache files to avoid cache collision or malicious access to cache, if a resource was deleted, it is possible to recreate resources with the same name. If a potentially attacker can create resources with the same name, they can access the original blob content for example with break-the-glass. I don't say it's highly probable, but most intrusions exploit a series of issues (in this case, they can create resources in that namespace somehow).

And of course it's not just security, but we don't want to reuse a different cache file on the first iteration just because we have `foo-bar.tar.gz` on the filesystem.

yitsushi (approved):

yitsushi:
Added some notes, but the doc itself looks good to me.

docs/adr/0003-workspace-blob-caching.md
chanwit: > If the controller container was restarted while the source was deleted (for example the controller crashed or rescheduled on a different node), it will not able to retrieve blob cache or blob artifact data from source.

yep, we'll adopt a PVC mechanism similar to the Source Controller for the future implementation.

yitsushi (dismissed):

squaremo (changes requested):
This ADR needs to be precise about the problem it addresses and how it addresses it. At present it gives its rationale in terms of adjectives -- "streamlined", "efficient", "optimized" -- but

 1. it's not clear why optimising is important, except in general terms
 2. the claims aren't supported by the explanation of the design

I suggest you revisit the context, and explain what is the "resource deletion problem" and what effect it has for users. Then you can rephrase the explanation of the design in terms of how it breaks down that problem, mechanically.

docs/adr/0003-workspace-blob-caching.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Decision
+
+1. **BLOB Creation and Storage**
+   * A gRPC function named `CreateWorkspaceBlob` will be invoked by the TF-Controller 
+     to compress the Workspace file system into a tar.gz format, which is then retrieved
+     as a byte array.
+   * The caching mechanism will be executed right before the Terraform Initialization step, ensuring that the latest and most relevant data is used.
+   * Each Workspace Blob will be cached on the TF-Controller's local disk, using the UUID of the Terraform object as the filename,`${uuid}.tar.gz`.
+   * To prevent unauthorized access to the cache entries, and cache collisions, the cache file will be deleted after the finalization process is complete.
```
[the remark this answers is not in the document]
yitsushi: That (and the previous line) prevents the scenario I described earlier.

yitsushi (dismissed):

docs/adr/0003-workspace-blob-caching.md
```diff
@@ -0,0 +1,55 @@
...
+5. **First MVP & Future Planning**
+   * For the initial MVP, the default pod local volume will be used.
+   * Since a controller restart will erase the BLOB cache, it's essential to maintain data integrity and availability. 
+     Consideration for using persistent volumes should be made for subsequent versions.
+
+## Consequence
+
+1. With the implementation of this architecture:
+   * BLOB management in TF-Controller will be optimized, leading to a more efficient and streamlined Terraform finalization process.
+   * Security measures will ensure the safety of the stored BLOBs, minimizing potential threats.
```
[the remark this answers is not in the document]
chanwit: > the added parts will be secure, but there's no overall change
yes.

yitsushi (approved):

docs/adr/0003-workspace-blob-caching.md
```diff
@@ -0,0 +1,56 @@
...
+The TF-Controller currently faces challenges related to the deletion of Terraform resources.
+These problems span across three categories:
+
+1. Single object deletion,
+2. Resources with dependencies deletion, and
+3. Namespace deletion.
+
+These problems must be fixed in the above order as (2) and (3) require single object deletion to be resolved first.
+
+Deleting a single TF object can sometimes be obstructed because it's tied to other resources like Source objects, Secrets, and ConfigMaps. If we try to remove it without deleting these resources, the TF object gets stuck in an inconsistent state, making it harder for users to manage their infrastructure smoothly.
```
squaremo, line 19: Based on the solution given below I would expect the problem to be that a Terraform object can't be finalized if resources it depends on are deleted _before_ it is. The description here suggests vice versa: that deleting a Terraform object before the things it relies on leads to the problem. Have I misunderstood?

docs/adr/0003-workspace-blob-caching.md
```diff
@@ -0,0 +1,56 @@
...
+Therefore, the TF-Controller is being enhanced to address this problem more efficiently, using the contents of generated Workspace BLOBs. Each BLOB contains all necessary information from the associated Source, Secrets, and ConfigMaps to ensure that TF-Controller finalization procedures can delete objects correctly.
+
+Currently, the TF-Controller downloads a Source BLOB and pushes it to a tf-runner. The tf-runner processes this BLOB to create a Workspace file system. It generates a backend configuration file, variable files, and other necessary files for the Workspace file system, using data from associated Secrets and ConfigMaps. This newly created Workspace file system is then compressed, sent back to the TF-Controller, and stored as a Workspace BLOB in the controller's storage.
```
squaremo, line 22: >  This newly created Workspace file system is then compressed, sent back to the TF-Controller, and stored as a Workspace BLOB in the controller's storage.

This is given as part of the solution below -- "A gRPC function named `CreateWorkspaceBlob` will be invoked by the TF-Controller to tell tf-runner to compress the Workspace file system into a tar.gz BLOB, which is then retrieved back to the controller". Is it current behaviour, or proposed behaviour?

docs/adr/0003-workspace-blob-caching.md
```diff
@@ -0,0 +1,56 @@
...
+These problems span across three categories:
+
+1. Single object deletion,
+2. Resources with dependencies deletion, and
+3. Namespace deletion.
+
+These problems must be fixed in the above order as (2) and (3) require single object deletion to be resolved first.
+
+Deleting a single TF object can sometimes be obstructed because it's tied to other resources like Source objects, Secrets, and ConfigMaps. If we try to remove it without deleting these resources, the TF object gets stuck in an inconsistent state, making it harder for users to manage their infrastructure smoothly.
+Therefore, the TF-Controller is being enhanced to address this problem more efficiently, using the contents of generated Workspace BLOBs. Each BLOB contains all necessary information from the associated Source, Secrets, and ConfigMaps to ensure that TF-Controller finalization procedures can delete objects correctly.
```
squaremo, line 20: What is it more efficient than?

If I'm a user that has got into this situation of having an undeletable Terraform object, is there anything I can do to unstick it -- presumably, recreating the missing resources would work, right? If so, I think you should mention that here or as part of the explanation of the problem; and, say that the proposed solution makes it much less likely that a user will have to step in and fix things themselves. (That really would be more efficient!)

squaremo (approved):
Thanks for adding the explanation of the problem. The text reads better as an argument, now -- here's the problem, we're going to solve it this way, because ...".

I was tripped up by the sentence "If we try to remove [a Terraform object] without deleting these resources [it relies on], the TF object gets stuck in an inconsistent state [...]" -- but I may have misunderstood. Revise it if you can see what I mean in my comment :-)

chanwit:
Will do a follow-up PR to address your comments.

PR #1039, opened 2023-10-10
fix: Tiltfile

yitsushi:


Makefile
```diff
@@ -12,6 +12,7 @@ BUILD_VERSION ?= $(shell git describe --tags $$(git rev-list --tags --max-count=
 # - .github/workflows/build-and-publish.yaml
 # - .github/workflows/release-runners.yaml
 # - .github/workflows/release.yaml
+# - Tiltfile
 LIBCRYTO_VERSION ?= 3.1.3-r0
```
yiannistri, line 16:
```suggestion
LIBCRYPTO_VERSION ?= 3.1.3-r0
```

yiannistri (commented):
I'd prefer to fix the typo in the Makefile, and update the Tiltfile accordingly.

Makefile
yitsushi: I hate dyslexia so hard to catch those. Thank you :)

yiannistri (approved):

PR #1042, opened 2023-10-11
Set runner.InitRequest.Upgrade to false by default

hmonsalv:
Prevents terraform runner to upgrade all providers, and will respect the definition of the terraform lock file if any. If required, this could be parametrized, so that the user can decide whether to enable upgrade option or not.

### DIFF
--- a/controllers/tf_controller_backend.go
+++ b/controllers/tf_controller_backend.go
@@ -344,7 +344,7 @@ terraform {
 
 	initRequest := &runner.InitRequest{
 		TfInstance: tfInstance,
-		Upgrade:    true,
+		Upgrade:    false,
 		ForceCopy:  true,
 		// Terraform:  terraformBytes,
 	}


chanwit:
@hmonsalv thank you for this PR.

Would you like to try adding a new field to the CR?
Maybe like this:

```yaml
kind: Terraform
spec:
  upgrade: false # default to true
```

The new `upgrade` flag would still be `true` by default to retrain compatibility. We would come back later for example in `v1alpha3` or `v1beta1` to set this value to false by default.

chanwit:
Here's another field, `RefreshBeforeApply`, which providers a flag but for the apply phase: https://github.com/weaveworks/tf-controller/blob/main/api/v1alpha2/terraform_types.go#L214

chanwit:
Make sense!

If many agreed, we could introduce this as a breaking change + docs by having the upgrade field default to false.

Then if users wanted the old behavior, they would go setting it to true.

@defreng would you be able to help with the docs part for this PR?

yitsushi:
Can you add the user configurable field in the spec? As far as I see the correct behaviour would be to NOT upgrade by default. The whole purpose of the controller is to maintain consistency with the git repository (gitops) and prevent most events that would result deviation from the source.

(also docs 😆)

daniel-ciaglia:
This feature is now present (https://github.com/flux-iac/tofu-controller/pull/1471) and this PR can be closed.

akselleirv:
Thanks @daniel-ciaglia !

PR #1061, opened 2023-10-20
feat: add remediation retry

### LINKED ISSUE #870: Implement the concept of remediation retry
Similar to what Flux has in the HelmRelease objects.

https://github.com/fluxcd/helm-controller/issues/41

yitsushi:
If `.Spec.Remediation` is not defined, it will retry indefinitely. If `.Spec.Remediation.Retries` is defined, it will stop re-queuing after
  it reached the limit.

```
❯ date && kubectl logs -n flux-system \
    --selector='app.kubernetes.io/name=tf-controller' \
    --tail=5000 --since-time="2023-10-25T13:08:00.000Z" \
  | grep -iE '(Reconciliation failed, retry)|(reached maximum number)|(Reset reconciliation failures count)' \
  | jq --raw-output '{"ts": .ts, "msg": .msg}'
Wed Oct 25 03:11:29 PM CEST 2023
{
  "ts": "2023-10-25T13:08:19.185Z",
  "msg": "Reset reconciliation failures count. Reason: resource changed"
}
{
  "ts": "2023-10-25T13:09:26.615Z",
  "msg": "Reconciliation failed, retry (1/3) after 5s. Generation: 7"
}
{
  "ts": "2023-10-25T13:09:58.774Z",
  "msg": "Reconciliation failed, retry (2/3) after 5s. Generation: 7"
}
{
  "ts": "2023-10-25T13:10:30.932Z",
  "msg": "Reconciliation failed, retry (3/3) after 5s. Generation: 7"
}
{
  "ts": "2023-10-25T13:10:40.948Z",
  "msg": "Resource reached maximum number of retries (3/3). Generation: 7"
}

❯ kubectl get terraforms.infra.contrib.fluxcd.io -n terraform helloworld-tf -o yaml | grep -A 2 '^ \+remediation'
  remediation:
    retries: 3
  retryInterval: 5s
```

Resolves #870


yitsushi:
Seems blocked on https://github.com/weaveworks/tf-controller/pull/1064

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		return createdHelloWorldTF.Status
+	}, timeout, interval).Should(Equal(expected))
+
+	// It should never reach 4 when Retry is set to 3.
+	expected = failedStatusCheckResult{
+		Retries: 4,
+	}
+	time.Sleep(15 * time.Second)
+	g.Eventually(func() interface{} {
```
squaremo, line 164: If the code is wrong, it could easily be wrong in such a way that it will keep retrying a few times over fifteen seconds, and the count will be higher than 4 the next time it's examined. In other words, this test is likely to pass when you want it to fail. Secondarily: this test will always take more than fifteen seconds to run.

The property under test here is that the system is marked as being in a failed state _if and only if_ the number of failures equals `.Spec.Remediation.Retries`. One direction is demonstrated above: failed state -> number of retries. The other direction is number of retries -> failed state. To show this, you could run the reconciler after it reaches the failure state*, and check that it exits without retrying. These aren't watertight proofs, by any stretch, but they are good enough.

*Since you want to run it _at least_ once, you can just call `r.Reconcile(...)`.

squaremo (changes requested):
The API change looks right and the general scheme of exiting Reconcile when the retries are exhausted is sensible.

I consider these things blockers:

 - I reckon ShouldRetry() is incorrect
 - possible NPE at tf_controller.go#L314

I had a hard time validating the logic for when to exit Reconcile due to retries. Specifically, ShouldRetry() is consulted both before and after having changed the value of .Status.ReconciliationFailures. I'd expect it to be checked _either_ before running anything ("hey, we're here again, but you've run out of goes") _or_ when it's exiting ("no more goes until something changes"). It could still be correct, but it's less clear how it's supposed to work -- are those additional check necessary? (If so, a comment explaining why would help me grok it, at least).

Before looking at that though, I think it's worth strengthening the tests (see comments) so changes can be made with extra confidence.

controllers/tf_controller.go
```diff
@@ -295,6 +295,28 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			log.Error(err, "unable to update status after planing")
+			return ctrl.Result{Requeue: true}, err
+		}
+	}
+
+	if !terraform.ShouldRetry() {
+		log.Info(fmt.Sprintf(
+			"Resource reached maximum number of retries (%d/%d). Generation: %d",
+			terraform.GetReconciliationFailures(),
+			terraform.Spec.Remediation.Retries,
```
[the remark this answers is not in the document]
yitsushi: True, originally it was in an extra if block.

// TODO

api/v1alpha2/terraform_types.go
```diff
@@ -890,6 +908,34 @@ func (in *Terraform) GetRunnerHostname(target string, clusterDomain string) stri
...
+func (in *Terraform) ResetReconciliationFailures() {
+	in.Status.ReconciliationFailures = 0
+}
+
+func (in *Terraform) ShouldRetry() bool {
+	if in.Spec.Remediation == nil {
+		return true
+	}
+
+	return in.GetReconciliationFailures() < in.Spec.Remediation.Retries
```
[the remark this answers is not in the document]
yitsushi: correct.

// TODO

charts/tf-controller/templates/crds.yaml
```diff
@@ -5269,6 +5269,17 @@ spec:
...
+                  the reconciliation fails. The default is to not perform any action.
+                properties:
+                  retries:
+                    description: Retries is the number of retries that should be attempted
+                      on failures before bailing. Defaults to '0', a negative integer
+                      equals to unlimited retries.
+                    format: int64
+                    type: integer
+                type: object
               retryInterval:
```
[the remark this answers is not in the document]
yitsushi: Yes I said the same, but otherwise we would have to duplicate that field, check if either of them is set, use that value, set priorities if both set, etc.

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,210 @@
...
+	g.Expect(err).ToNot(HaveOccurred())
+
+	It("should be created and attached successfully.")
+	g.Expect(k8sClient.Create(ctx, &helloWorldTF)).Should(Succeed())
+	defer waitResourceToBeDelete(g, &helloWorldTF)
+
+	By("checking that the TF resource existed inside the cluster.")
+	helloWorldTFKey := types.NamespacedName{Namespace: "flux-system", Name: terraformName}
+	createdHelloWorldTF := infrav1.Terraform{}
+	g.Eventually(func() bool {
```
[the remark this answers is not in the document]
yitsushi: That's a fully copy paste from a different test and I added extra tests at the end end removed unrelated tests, but mostly kept as it is, and that's why I said a few months ago to not address it per test, but create a follow-up issue, refractor tests, otherwise we will end up with the same problems if anyone tries to create a new test case as everyone will start their new test case by finding one that "kind of tests" the same thing, and change it to fulfil the needs of the test case.

controllers/tf_controller.go
```diff
@@ -295,6 +295,28 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	}
 
+	// Reset retry count if necessary.
+	revisionChanged := sourceObj.GetArtifact().Revision != terraform.Status.LastAttemptedRevision
+	generationChanges := terraform.Generation != terraform.Status.ObservedGeneration
+	if revisionChanged || generationChanges {
+		log.Info("Reset reconciliation failures count. Reason: resource changed")
+		terraform.ResetReconciliationFailures()
+		if err := r.patchStatus(ctx, req.NamespacedName, terraform.Status); err != nil {
+			log.Error(err, "unable to update status after planing")
```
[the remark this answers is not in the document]
yitsushi: It feels like a copy paste, so I assume we have a "unable to update status after planing" somewhere else in the code too.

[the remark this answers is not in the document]
yitsushi: Yes, more than one: https://github.com/search?q=repo%3Aweaveworks%2Ftf-controller+planing&type=code

yitsushi (commented):
I read the comments (most of them), made some comments (some for myself) so I remember tomorrow. That's where my brain said "enough".

controllers/tf_controller.go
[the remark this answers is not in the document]
yitsushi: It's not a NPR.
`ShouldRetry()` will return true if `in.Spec.Remediation == nil`, because of that, if `!ShouldRetry()` is true, the value is not `nil`.

```go
func (in *Terraform) ShouldRetry() bool {
	if in.Spec.Remediation == nil {
		return true
	}
	// ...
```

I'll add an extra check just in case, but now it's not NPR.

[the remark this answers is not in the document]
yitsushi: Added a comment about it to make is clear, but I feel testing it redundant.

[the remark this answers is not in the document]
squaremo: You are right, I got myself in a twist there. Best not to rely on it, though.

api/v1alpha2/terraform_types.go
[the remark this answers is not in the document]
yitsushi: fixed in 1b9c6a58f2968e5f3083c189a03fd7df9e03ceb3

squaremo (commented):
Using the condition makes the first part of the test more reliable :+1: but I think there's still a problem with the second part -- details in the commentary.

controllers/tc000243_remediation_retry_test.go
yitsushi: Originally I just made a simple `Get` call, but I tend to put **everything** (that contacts kube api) in an `Eventually` block in case the `Get` call returns with an error and causing an intermittent/flaky test failure.

I can see it can pass the check in a faulty situation.

How about this:
```

recheckHelloWorldTF := infrav1.Terraform{}
g.Eventually(func() interface{} {
		return k8sClient.Get(ctx, helloWorldTFKey, &recheckHelloWorldTF)
	}, timeout, interval).Should(gomega.Succeed()))
```
and than check if `recheckHelloWorldTF` and `createdHelloWorldTF` has the same properties/values.

I really don't want to remove that `Eventually`, we had to fix a lot of missing `Eventually` calls just because the API didn't like the result of a d20 roll of the CI at that particular time of the day.

api/v1alpha2/terraform_types.go
```diff
@@ -826,6 +845,30 @@ func TerraformStateLocked(terraform Terraform, lockID, message string) Terraform
 	return terraform
 }
 
+// TerraformReachedLimit will set a new condition on the Terraform resource
+// indicating that the resource has reached its retry limit.
+func TerraformReachedLimit(terraform Terraform) Terraform {
+	newCondition := metav1.Condition{
+		Type:    ConditionRetryLimitReached,
```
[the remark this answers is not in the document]
yitsushi: First I used Stalled and then somewhere I found an example to use `Status: ConditionFalse` for retry resets, and then I found a different one that used custom type, but remove after reset and... well I ended up with a mix of them it seems.

And it was easier to just check that specific Status in the Conditions slice, rather than checking if it's Stalled and has given reason (not it's not really, just an extra `&& status.Reason == infra.XXX`, but feels like it's more complex).

▶️ I'll change it to Stalled and move the reason into a constant.

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,210 @@
...
+		Retries int64
+	}
+	expected := failedStatusCheckResult{
+		Type:    "Ready",
+		Reason:  "ArtifactFailed",
+		Message: "rpc error: code = Unknown desc = failed to untar artifact, error: requires gzip-compressed body: gzip: invalid header",
+		Status:  metav1.ConditionFalse,
+		Retries: 3,
+	}
+	g.Eventually(func() interface{} {
```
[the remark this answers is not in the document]
squaremo: I think you've addressed this by adding the `RetryLimitReached` (possibly soon to be `Stalled, Reason=RetryLimitReached` condition.

controllers/tc000243_remediation_retry_test.go
squaremo: > the `Get` call returns with an error and causing an intermittent/flaky test failure

Oh no, an error that's not `NotFound`? What a pain. Since `Eventually` is part of the assertion being made, I think it would be clearer to paper over the unreliable API by using https://pkg.go.dev/k8s.io/client-go/util/retry#OnError, wrapped in a helper. wdyt?

yitsushi: I don't really see the difference is a test, the `retry.OnError` would be even more robust just to make sure the test env doesn't :hankey: itself because of low resources:

**With `Eventually`:**
```go
g.Eventually(func() error {
	return k8sClient.Get(ctx, helloWorldTFKey, &recheckHelloWorldTF)
}, timeout, interval)..Should(Succeed())
```

**With `OnError`:**
```
// Note: we would have to define a wait.Backoff to make sure it similarly to
// Eventually elsewhere.
err := retry.OnError(retry.DefaultRetry, func(_ error) bool {
	// Technically we don't care.
	return true
}, func() error {
	return k8sClient.Get(ctx, helloWorldTFKey, &recheckHelloWorldTF)
})
g.Expect(err).To(Succeed())
```

If I simplify everything, it's still much more complex for no benefits.
```go
err := retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
	return k8sClient.Get(ctx, helloWorldTFKey, &recheckHelloWorldTF)
})
g.Expect(err).To(Succeed())
```

Sure, we can separate the retry logic from the actual call, but
1. We can check that in Eventually too but on in a fancy two-separate-function way.
2. Most of the time we don't care what went wrong, we just want to retry it until it successful or throw an error.

api/v1alpha2/terraform_types.go
```diff
@@ -900,6 +943,34 @@ func (in *Terraform) GetRunnerHostname(target string, clusterDomain string) stri
...
+
+func (in *Terraform) GetReconciliationFailures() int64 {
+	return in.Status.ReconciliationFailures
+}
+
+func (in *Terraform) IncrementReconciliationFailures() {
+	in.Status.ReconciliationFailures++
+}
+
+func (in *Terraform) ResetReconciliationFailures() {
```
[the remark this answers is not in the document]
yitsushi: changed it to be a private function (non-exported).

controllers/tc000243_remediation_retry_test.go
yitsushi: An option couldn't be to create a global wrapper around all `Get` and use that in all tests, but than either we create a new package or copy paste the wrapper function all over to all modules and it wouldn't help much in most cases where `Eventually` test things and not just a safeguard.

squaremo: I think

```
var recheckHelloWorldTF infrav1.Terraform
g.Expect(reliableGet(ctx, helloWorldTFKey, &recheckHelloWorldTF)).To(Succeed())
```

is easier to read than

```
var recheckHelloWorldTF infrav1.Terraform
g.Eventually(func() error {
	return k8sClient.Get(ctx, helloWorldTFKey, &recheckHelloWorldTF)
}, timeout, interval).Should(Succeed())
```

 -- it's shorter, but also there's no `Eventually` to make me think that there's more going on than there really is.

> It couldn't be an option to create a global wrapper around all Get and use that in all tests

Why not?

squaremo: Proportionality check: I am not fussed about this particular point, if the tests are right.

yitsushi: In general I agree with that, but I wouldn't create a function just because we have 3 instances where we would use it, but I'll keep in mind for example when we get to refactor all the tests to remove unnecessary parts and simplify things because we can.

controllers/tc000243_remediation_retry_test.go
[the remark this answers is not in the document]
squaremo: I filed https://github.com/weaveworks/tf-controller/issues/1130 for Eventually reform.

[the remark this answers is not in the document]
yitsushi: Issue about the copy paste mess with sus checks: https://github.com/weaveworks/tf-controller/issues/1131

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,210 @@
...
+
+	It("should be reconciled and contain some status conditions.")
+	By("checking that the TF resource's status conditions has 1 element.")
+	g.Eventually(func() int {
+		err := k8sClient.Get(ctx, helloWorldTFKey, &createdHelloWorldTF)
+		if err != nil {
+			return -1
+		}
+		return len(createdHelloWorldTF.Status.Conditions)
+	}, timeout, interval).Should(Equal(1))
```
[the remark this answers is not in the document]
squaremo: I think this would be covered by #1131.

controllers/tc000243_remediation_retry_test.go
squaremo: > If I replace `time.Sleep(...)` with `reconciler.Reconcile(...)`, then the test fails; it's _relying_ on not being requeued. But a requeue can happen without being triggered by an update; e.g., every [cache.SyncPeriod](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.16.3/pkg/cache/cache.go#L166).
> 
> To avoid retrying without something changing, you may need to check the generation against the `.observedGeneration` field of the `RetryLimitReached` condition*, and make sure that latter field is set.

This is the last thing to address. My diagnosis and my suggestion above are a ways wrong, now I read them back. I suggested that it should check the observed generation _of the Stalled condition_ -- but that's not right, because what matters is if the spec changed since the last time it tried, whatever happened. The code here already checks the (`.metadata.generation != .status.observedGeneration`](https://github.com/weaveworks/tf-controller/blob/870-retry/controllers/tf_controller.go#L300), which is right.

This is how it appears to work, in pseudo-code:

```
if changed since last time: # L299
  reset failure counter
if failure counter >= retry limit: # L441
  set stalled condition and exit
run program
if program failed: # L467
  increment retry counter
  requeue and exit # L493
else:
  reset failure count
requeue and exit
```
This is correct, on the face of it.

In the test case, once the resource has reached the retry limit, the test sleeps for 15s, then checks it hasn't run again. According to the pseudo-code above (and I believe the code too) that will succeed trivially, because once it's reached the limit, it doesn't requeue it.

However, a resource _can_ be queued in this state, e.g., if the client cache is "synced" (where controller-runtime runs everything in the client cache through the reconciler, to flush out any missed changes). This won't happen in the test, because the sync period is ten hours -- but it would happen, more than twice a day, in practice. The controller should behave according to the pseudo-code above, when it does happen.

If I replace the `time.Sleep()` with `reconciler.Reconcile()`, it simulates the resource being queued arbitrarily. The test now **fails**, because the stalled condition is no longer present. That's surprising! ...given the apparent correspondence of the code and the scheme given above.

I added some printlns to where it decides if the object has changed. It turns out that when it is run _after_ hitting the retry limit, `.status.lastAttemptedRevision` is empty; that is compared to the source revision (L299), which is not empty, and the controller decides it's changed and resets the failure count. So, long story short: generally the code fits the pseudo-code above, but there's a bug in how the status is patched when it hits the retry limit.

yitsushi: Added the check to the test run it (without other code changes) and indeed `lastAttemptedRevision` was empty.

I know it does not fix the cache sync issue in test, but this should fix the reset without `lastAttemptedRevision`: https://github.com/weaveworks/tf-controller/pull/1061/commits/935bf5bff8f7dffb4d5ada588022c130c9cea55f ?

squaremo: tf_controller.go L451: When `!ShouldRetry()`, `terraform.TerraformReachedLimit()` is called, and that in turn [calls `SetTerraformReadiness`](https://github.com/weaveworks/tf-controller/pull/1061/files#diff-3f0c6c788c08365729a653db804a723d7755bd4ff5016186f527223e4d4c46d9R858) with an empty string for the last attempted revision.

yitsushi: :woman_facepalming: good catch.

yitsushi: I don't even know why I called `SetTerraformReadiness` there.

.github/workflows/helm-test.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
           python-version: "3.10"
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@e8788873172cb653a90ca2e819d79d65a66d4e76 # v2.4.0
+        uses: helm/chart-testing-action@e6669bcd63d7cb57cb4380c33043eebe5d111992 # v2.6.1
```
yitsushi, line 31: Will be removed once https://github.com/weaveworks/tf-controller/pull/1147 is merged.

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,225 @@
...
+	sourcev1 "github.com/fluxcd/source-controller/api/v1"
+	infrav1 "github.com/weaveworks/tf-controller/api/v1alpha2"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// +kubebuilder:docs-gen:collapse=Imports
+
+func Test_000243_remediation_retry_test(t *testing.T) {
+	Spec("This spec describes the behaviour of a Terraform resource that obtains a bad tar.gz file blob from its Source reference.")
```
squaremo, line 22:
```suggestion
	Spec("This spec describes the retry behaviour of a Terraform resource that repeatedly fails.")
```

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,225 @@
...
+					Message: c.Message,
+					Status:  c.Status,
+				}
+			}
+		}
+		return createdHelloWorldTF.Status
+	}, timeout, interval).Should(Equal(expected))
+	g.Expect(createdHelloWorldTF.Status.ReconciliationFailures).To(Equal(int64(3)))
+
+	reconcileResult, err := reconciler.Reconcile(ctx, controllerruntime.Request{
```
squaremo, line 157: :+1: nice idea, testing the response and error.

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,225 @@
...
+	g.Expect(err).To(Succeed())
+	g.Expect(reconcileResult.Requeue).To(BeFalse())
+
+	recheckHelloWorldTF := infrav1.Terraform{}
+	g.Eventually(func() interface{} {
+		return k8sClient.Get(ctx, helloWorldTFKey, &recheckHelloWorldTF)
+	}, timeout, interval).Should(Succeed())
+
+	var retryCondition *metav1.Condition
+	for _, cond := range recheckHelloWorldTF.Status.Conditions {
```
squaremo, line 172: [`apimeta.FindStatusCondition`](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/meta#FindStatusCondition) will do this bit for you (and you would have to check that it's stalled for the right reason).

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,225 @@
...
+		if cond.Type == meta.StalledCondition && cond.Reason == infrav1.RetryLimitReachedReason {
+			originalRetryCondition = &cond
+			break
+		}
+	}
+
+	g.Expect(retryCondition).ToNot(BeNil(), "Terraform resource should have retry limit reached status condition")
+	g.Expect(originalRetryCondition).ToNot(BeNil(), "Terraform resource should have retry limit reached status condition")
+	g.Expect(recheckHelloWorldTF.Status.ReconciliationFailures).To(Equal(createdHelloWorldTF.Status.ReconciliationFailures))
+	g.Expect(retryCondition.ObservedGeneration).To(Equal(originalRetryCondition.ObservedGeneration))
```
squaremo, line 189: These will probably both be `0`, and trivially succeed -- worth checking.

controllers/tc000243_remediation_retry_test.go
```diff
@@ -0,0 +1,225 @@
...
+				return failedStatusCheckResult{
+					Type:    c.Type,
+					Reason:  c.Reason,
+					Message: c.Message,
+					Status:  c.Status,
+					Retries: createdHelloWorldTF.Status.ReconciliationFailures,
+				}
+			}
+		}
+		return createdHelloWorldTF.Status
```
squaremo, line 223: Returning the `.Status` muddies the water a bit -- it's a value that fails the loop, but it's not `nil`, so ... why return it? (I'll presume copypasta)

squaremo (dismissed):
Looks in order, thanks for persevering @yitsushi :star:
One thing that needs addressing: a newly generated deepcopy method has not been committed (see the copy on terraform_types.go).

api/v1alpha2/terraform_types.go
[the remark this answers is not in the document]
yitsushi: It does not do that to on my side :O I'll check tomorrow.

controllers/tc000243_remediation_retry_test.go
yitsushi: Yes, it should be `nil` here as well, no reason to return with any values.

api/v1alpha2/terraform_types.go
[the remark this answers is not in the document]
yitsushi: nvm, recently I used `go test` to speed up re-runs

squaremo (approved):
Thanks for sticking at it @yitsushi, this looks good. I made a note about checking restarts when there's a new source revision -- you don't need to act on that (or to include the suggestion in a commit). The big green button awaits!

PR #1075, opened 2023-10-30
chore(deps): Update google.golang.org/grpc and golang.org/x/net

yitsushi:
Reason: https://github.com/advisories/GHSA-m425-mq94-257g

> Impact
>
> In affected releases of gRPC-Go, it is possible for an attacker to
> send HTTP/2 requests, cancel them, and send subsequent requests, which
> is valid by the HTTP/2 protocol, but would cause the gRPC-Go server to
> launch more concurrent method handlers than the configured maximum
> stream limit.
>
> Patches
>
> This vulnerability was addressed by #6703 and has been included in
> patch releases: 1.56.3, 1.57.1, 1.58.3. It is also included in the
> latest release, 1.59.0.
>
> Along with applying the patch, users should also ensure they are using
> the grpc.MaxConcurrentStreams server option to apply a limit to the
> server's resources used for any single connection.
>
> Workarounds
>
> None.

References:
* https://github.com/advisories/GHSA-m425-mq94-257g
* https://github.com/grpc/grpc-go/pull/6703


go.mod
```diff
@@ -205,7 +197,7 @@ replace (
 	github.com/hashicorp/terraform-exec v0.16.1 => github.com/tf-controller/terraform-exec v0.15.1-0.20220809152546-4850a69faedb
 
 	// Fix CVE-2023-32731
-	google.golang.org/grpc => google.golang.org/grpc v1.53.0
+	google.golang.org/grpc => google.golang.org/grpc v1.56.3
```
yiannistri, line 200: I would move this out of the replace directive, as I don't understand why we need to pin this to an earlier version. Alternatively if we *have* to pin it to a specific version, I would updated the comment above to indicate which CVE this addresses.

yiannistri (approved):

PR #1076, opened 2023-10-30
Avoid drift detection when using manual mode

### LINKED ISSUE #890: The controller should not perform drift detection if it's triggered by a planning request
Manual approval is enabled, and the initial plan is approved and applied successfully.

Now the terraform code changes a little bit, it requires a small update. But I cannot find a new plan ID. Then how to approve the update after the initial plan is applied? I don't find this in the documentation either.

luizbafilho:
closes #890

To fix https://github.com/weaveworks/tf-controller/issues/890, this PR implements @chanwit proposal.

- This avoids running drift detection when using manual mode.


yiannistri:
We have decided to close this PR without merging it as it is changing existing behavior that users _may_ not anticipate. We will go back to the drawing board about how to best address this issue.

PR #489, opened 2023-01-04
Feat/add lock timeout option

emalihin:

PR #1087, opened 2023-11-01
Add support for Terraform `LockTimeout` feature in the plan stage

### LINKED ISSUE #482: non-locking plans
Hello,

There are scenarios where we want to run the same state plan from multiple k8s clusters, for example:
* All dev->prod clusters create/manage the same Sentry project with the same config
* Multi-cluster scenario where there are two or more prod clusters manage the same infra with the same config

However we sometimes run into state-lock errors, where timings of plans clash, and plan that started second cannot run due to the first plan locking the state.

What do you think about adding an option to set `-lock=false` on the plan stage?

Nalum:
Add new `LockTimeout` field to the Terraform API. See [Terraform Docs](https://developer.hashicorp.com/terraform/cli/commands/plan#lock-timeout-duration) for details on this flag.

Thanks to @emalihin for the work on #489.

This PR also updates the `Makefile` and the docker files changing the `go build` commands to run against the directory rather than a specific file e.g. `go build ./cmd/tfctl` instead of `go build cmd/tfctl/main.go`.

I've also updated the `Makefile` so that the default target is `help`.

Closes #482


Makefile
```diff
@@ -1,4 +1,4 @@
-
+.DEFAULT_GOAL := help
```
yitsushi, line 1: ❤️

api/v1alpha2/terraform_types.go
```diff
@@ -479,6 +479,14 @@ type TFStateSpec struct {
 	//
 	// +optional
 	LockIdentifier string `json:"lockIdentifier,omitempty"`
+
+	// LockTimeout is a Duration string that instructs Terraform to retry acquiring a lock for the specified period of
+	// time before returning an error. The duration syntax is a number followed by a time unit letter, such as `3s` for
+	// three seconds.
```
yitsushi, line 485: [not a blocker]
Can we add a note about the `0s` value, I think it's not obvious what happens if the timeout is `0s`.

yitsushi (approved):
lgtm 👍 

(I really like all the replacements from `main.go` file references to `./dir`)

api/v1alpha2/terraform_types.go
Nalum: Added extra info :+1:

yitsushi (approved):

PR #1093, opened 2023-11-02
Variable overriding

kuritka:
The purpose of this change is to enable overriding variables mechanism. I want to define default values (specified in `varsFrom` via configMap or Secret) and customize them via `vars`.

All I do is just change the order of writing to the variables 

Example:
create a configMap named `defaults-cm` with the value `costCenter=123456` and add it to all Terraform manifests.
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: default-cm
data:
  costCenter: 123456
```

Everything that tf-runner will run will have the default variable `costCenter=123456`, because all Terraform Manifests have the field:
```yaml
spec:
  varsFrom:
    - kind: configMap
      name: defaults-cm
```

A request is received that the part of the infrastructure that is specified by the Terraform  manifest named `override` has `costCenter=888888`. I can do this easily, because I override the value via vars:

```yaml
kind: Terraform
Metadata:
  name: override
spec:
  vars:
  - name: costCenter 
    value: 88888
```

Thanks to this functionality I can more easily customize the entire infrastructure.


squaremo:
> Hi @LappleApple , do you have any document in the style of CONTRIBUTION.md. I see that `TestMain(m *testing.M)` requires some pre-requisites. To be able to run tests locally.

Does https://github.com/weaveworks/tf-controller/blob/main/DEVELOPMENT.md help?

PR #1096, opened 2023-11-03
Vars overriden by VarsFrom

### LINKED ISSUE #634: Using nested variable blocks from a configmap
Hi,

I'm trying to use a module that requires variables with maps, lists ...
Let's say that I want to use this [irsa-eks](https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/examples/iam-role-for-service-accounts-eks/main.tf#L71) module.
How would I define the variables into a configmap please?
The following code won't work obviously

```yaml
apiVersion: v1
data:
  role_name: "cert-manager"
  attach_load_balancer_controller_policy: true
  oidc_providers:
    ex:
      provider_arn: ${oidc_cluster_issuer}
      namespace_service_accounts: ["kube-system:alb-controller"]
  tags:
    env: ${environment}
    cluster: ${cluster_name}
kind: ConfigMap
metadata:
  name: tfvars-dev-hiving-hen
  namespace: flux-system
```

Thank you

kuritka:
Fixes #634 

_I apologize for accidentally closing the [previous PR](https://github.com/weaveworks/tf-controller/pull/1093) I did `git reset -hard <start commit> and git push -f` on my fork which automatically closed the upstream PR.  It behaved a bit differently than I expected._ 

---
The purpose of this change is to enable overriding variables mechanism. I want to define default values (specified in `varsFrom` via configMap or Secret) and customize them via `vars`.

All I do is just change the order of writing to the variables + update overriding in 090xxx test.

Example:
create a configMap named `defaults-cm` with the value `costCenter=123456` and add it to all Terraform manifests.
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: default-cm
data:
  costCenter: 123456
```

Everything that tf-runner will run will have the default variable `costCenter=123456`, because all Terraform Manifests have the field:
```yaml
spec:
  varsFrom:
    - kind: configMap
      name: defaults-cm
```

A request is received that the part of the infrastructure that is specified by the Terraform  manifest named `override` has `costCenter=888888`. I can do this easily, because I override the value via vars:

```yaml
kind: Terraform
Metadata:
  name: override
spec:
  vars:
  - name: costCenter
    value: 88888
```

Thanks to particular Terraform.Spec.Vars I can override globaly defined variables specified in Terraform.Spec.VarsFrom and easily customize the entire infrastructure.

### DIFF
--- a/controllers/tc000090_varsfrom_override_and_controlled_outputs_test.go
+++ b/controllers/tc000090_varsfrom_override_and_controlled_outputs_test.go
@@ -151,7 +151,7 @@ func Test_000090_varsfrom_override_and_controlled_outputs_test(t *testing.T) {
 	expectedOutputValue := map[string]string{
 		"Name":        "tf-output-" + terraformName,
 		"Namespace":   "flux-system",
-		"Value":       "Hello, my overridden cat!",
+		"Value":       "Hello, my cat!",
 		"OwnerRef[0]": string(createdHelloWorldTF.UID),
 	}
 	g.Eventually(func() (map[string]string, error) {
--- a/docs/use_tf_controller/to_set_variables_for_Terraform_resources.md
+++ b/docs/use_tf_controller/to_set_variables_for_Terraform_resources.md
@@ -27,8 +27,9 @@ Inline variables can be set using `vars`. The `varsFrom` field accepts a list of
 You may use the `varsKeys` property of `varsFrom` to select specific keys from the input or omit this field
 to select all keys from the input source.
 
-Note that in the case of the same variable key being passed multiple times, the controller will use
-the lattermost instance of the key passed to `varsFrom`.
+Values from `varsFrom` are always read first and then values from `vars` are read. If the same key is used in `varsFrom`
+as in `vars`, the value from `varsFrom` will be overwritten. This mechanism allows us to have the global variables set 
+once in one of the ConfigMaps or Secrets, and if necessary we can overwrite the value directly in the Terraform object.
 
 ```yaml hl_lines="15-20 22-28"
 apiVersion: infra.contrib.fluxcd.io/v1alpha2
--- a/runner/server_generate_tf_vars.go
+++ b/runner/server_generate_tf_vars.go
@@ -243,13 +243,6 @@ func (r *TerraformRunnerServer) GenerateVarsForTF(ctx context.Context, req *Gene
 		vars["values"] = &apiextensionsv1.JSON{Raw: buf.Bytes()}
 	}
 
-	log.Info("mapping the Spec.Vars")
-	if len(terraform.Spec.Vars) > 0 {
-		for _, v := range terraform.Spec.Vars {
-			vars[v.Name] = v.Value
-		}
-	}
-
 	log.Info("mapping the Spec.VarsFrom")
 	// varsFrom overwrite vars
 	for _, vf := range terraform.Spec.VarsFrom {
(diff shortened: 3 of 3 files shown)


yitsushi (commented):
Can we add a note about it here: https://github.com/weaveworks/tf-controller/blob/main/docs/use_tf_controller/to_set_variables_for_Terraform_resources.md?plain=1#L30

Now it says "the controller will use
the lattermost instance of the key passed to `varsFrom`" only, and I found no docs about the priority. This would be a good moment to add this info (".spec.vars takes precedent over varsFrom" or something like that)

kuritka:
@yitsushi, thx for info, I amended the change. Let me know if you like to do any changes

yitsushi (commented):
Thinking more about this, the reverse logic would make more sense to me in some cases. The question is, which one is more common.

### 1. Option (original logic)
- I create a Terraform resource with default variables as `vars`
- I create the variables secret and if something is not defined in there it falls back to the default `vars`

### 2. Option
- I create the values secret with all options
- I create one or more terraform objects using that "default value" secret and override values as `vars`.
- (this can be achieved with adding a new `varsFrom` with new values as it will use the value from the last definition of a key).

@chanwit what's your opinion.

k0da:
Our use case:
We manage fleet of dozens k8s clusters from this. Each `Terraform` object have varsFrom: global-vars (it defines environment, kubernetes version, etc). Now we need to start rolling out a new k8s upgrade. So we start with one object first, it overrides global-vars and performs upgrade. Once we confident enough we enable global (the rest) rollout by changing in global-vars reference. And remove individual overrides.

Right now. We have to temporarly remove varsFrom reference and copy every k/v pair from there into an object itself to be able to override single var.

yitsushi:
> Right now. We have to temporarly remove varsFrom reference and copy every k/v pair from there into an object itself to be able to override single var.

@k0da: With the current logic. You can create an "update-vars" secret and add to the end of the list of `fromVars`, it will override the value from the global.

That even looks more reasonable to me in this case as you can have a global config set everywhere  and an upgrade valuaes you just add to terraform resources and you can simply remove the secret reference when you updated the global, you get the same upgrade values that way on all "test" upgraded resources.

k0da:
@yitsushi I see, but this way you'll have to track additional resource. controller won't kick a reconcilation on actual configmap change. And are configmap/secret lists are sorted?

yitsushi:
> And are configmap/secret lists are sorted?

They are parsed in the order you define.

> Note that in the case of the same variable key being passed multiple times, the controller will use
> the lattermost instance of the key passed to `varsFrom`.
>
> https://github.com/weaveworks/tf-controller/blob/main/docs/use_tf_controller/to_set_variables_for_Terraform_resources.md?plain=1#L30-L31

> controller won't kick a reconcilation on actual configmap change

when you add it will reconcile and after that it will reconcile in the defined interval. So for example
1. define an upgrade config map
2. update the terrafrom object
   - add the new configmap as the last item in the list
   - change the interval to a lower value so you get faster feedback (or use `tfctl reconcile`)

I don't say that's how it should be done, but it is possible to achieve the same thing with the current logic, and if I would have to do that I would create one upgrade config and add to all resources I want to upgrade before changing global options.

k0da:
@yitsushi makes sense. Lets collect more feedback

yitsushi:
After a discussion with @chanwit, the current logic is intentional.

> vars are visible as plain values with an intent to be overridden by CMs or Secrets.
> values can be overridden by addition varFroms. It's promoting better security.

This PR would introduce a braking change and promote the use of vars over secrets which is something we should avoid. As I described above, the same effect can be achieved using the `varFroms` list by defining a different value set and put it at end end (or at least further down) the list.

As per the discussion with @chanwit, I'm closing this PR.

PR #1098, opened 2023-11-03
Fix break glass command; respect all flags

### LINKED ISSUE #1097: [bug/tfctl] kubectl context and other flags not respected on break-glass
**Plain error description**

If I run tfctl like `tfctl break-glass --context bbbb --namespace 1234 my-terraform-thing`, where `bbbb` is not the default context, I get an error like
```
Error from server (NotFound): pods "my-terraform-thing-tf-runner" not found
```

(If the cluster that is the default context does not have a namespace called `1234`, I get an error like
```
Error from server (NotFound): namespaces "1234" not found
```
-> I put this here if someone is searching for the error, like I did)

This (when I look at the problem I found) means that right now the break-glass shell is run on a completely different cluster. If the same Terraform object happens to be there, this will break-glass on the wrong pod.

**Investigation result**

This error comes from [this line](https://github.com/weaveworks/tf-controller/blob/0452bc9690872c7f841c10cc9c7f70d7278fe339/tfctl/break_glass.go#L112) in the break-glass logic, which does not respect any flag other than the namespace.

**Possible solutions**

I will open a PR that fixes this and similar problems shortly.

DrBu7cher:
fixes #1097

As I said in #1097, break-glass could easily target the wrong cluster. This happened to me.
This PR adds the flags to the command, that should be passed and can be passed according to `tfctl help`.

Please try to get it into 0.16.0, so the fix gets out fast.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -18,16 +18,19 @@ to merge them for you after reviews.
 
 ## Protobuf Setup
 
-TF-controller requires a specific version of Protobuf compiler and its Go plugins. 
+TF-controller requires a specific version of Protobuf compiler and its Go plugins.
 
 * Protoc: version [3.19.4](https://github.com/protocolbuffers/protobuf/releases/download/v3.19.4/protoc-3.19.4-linux-x86_64.zip)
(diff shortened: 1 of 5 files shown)


chanwit (approved):
LGTM!

Thank you @DrBu7cher for this PR! :1st_place_medal:

Issue #833, opened 2023-08-02
Cluster scoped roles and cluster-admin RoleBinding in reconciler


SebastianBakala-TomTom:
1. Is it possible to run controllers on non-ClusterScoped roles?

2. Does the reconciler really needs cluster-admin role?
https://github.com/weaveworks/tf-controller/blob/main/charts/tf-controller/templates/rbac.yaml#L170

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: tf-cluster-reconciler
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cluster-admin
subjects:
- kind: ServiceAccount
  name: {{ include "tf-controller.serviceAccountName" . }}
  namespace: {{ .Release.Namespace }}
```

chanwit:
Hi @SebastianBakala-TomTom 

1. Yes
2. No you don't really need it.

The RBAC provided is to provide the out-of-the-box experience. Please feel free to scope down the permissions. However, you have to make that your setup has enough permissions to talk to other Flux components which are Source and Notification Controllers.

SebastianBakala-TomTom:
Would you mind if I create a pull request with a fixed rbac configuration?

chanwit:
Please feel free to do so. Maybe adding an option to Helm Chart so that we fully have backward compatibility?

cc @squaremo as Michael is taking care of security-related stuff.

squaremo:
It should certainly not need `cluster-admin`. I think it probably needs
 - get/list/watch on Terraform objects and things they refer to
 - get/create on pods (or jobs? whichever the runner uses)

in a ClusterRole (so it can be applied to any namespace), and RoleBinding or a ClusterRoleBinding depending on whether it's watching a namespace or all namespaces. Does that sound about right? (thinking about RBAC still makes my brain go funny)

LappleApple:
@SebastianBakala-TomTom Heya, checking in to ask if you still want to make PR related to this issue?

SebastianBakala-TomTom:
Hello @LappleApple 
Sure I want. Gimme few hours

LappleApple:
@SebastianBakala-TomTom Take your time. Not going to be checking on this over the weekend.

PR #1099, opened 2023-11-05
ci: limit rbac for the reconciler

SebastianBakala-TomTom:
solves 2nd question from #833


yitsushi:
This PR will fix the helm-test workflow: https://github.com/weaveworks/tf-controller/pull/1147

SebastianBakala-TomTom:
Sorry @yitsushi. I missed your comment.

charts/tf-controller/templates/rbac.yaml
```diff
@@ -46,6 +46,69 @@ rules:
...
+  resources:
+  - jobs
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
```
yiannistri, line 95: Not sure if this is needed? I could find no Job object references in the code, could you explain why this was added?

yiannistri (commented):
Tested and works fine. Unsure about `jobs` RBAC though.

yiannistri (approved):
Tested with "plan and manual apply" as well as "plan with automatic apply" and worked fine. Secrets and configmaps are also created and got no RBAC errors in the logs.

Thank you for the PR @SebastianBakala-TomTom :1st_place_medal:

PR #1152, opened 2023-11-29
Add CreateWorkspaceBlob RPC method to Runner service

luizbafilho:
This RPC method will be used to create a compressed tarball of the .terraform directory of the runner.

### DIFF
--- a/internal/storage/storage.go
+++ b/internal/storage/storage.go
@@ -35,6 +35,8 @@ func ArchiveDir(dir string) (out string, err error) {
 	if err != nil {
 		return "", err
 	}
+	defer tf.Close()
+
 	tmpName := tf.Name()
 	defer func() {
 		if err != nil {
@@ -46,14 +48,17 @@ func ArchiveDir(dir string) (out string, err error) {
 	mw := io.MultiWriter(h, tf)
 
 	gw := gzip.NewWriter(mw)
+	defer gw.Close()
+
 	tw := tar.NewWriter(gw)
+	defer tw.Close()
+
 	if err := filepath.Walk(dir, func(p string, fi os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
 
-		// Ignore anything that is not a file (directories, symlinks)
-		if !fi.Mode().IsRegular() {
+		if fi.IsDir() {
 			return nil
 		}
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -25,6 +25,7 @@ COPY cmd/manager/ cmd/manager/
 COPY controllers/ controllers/
 COPY mtls/ mtls/
 COPY runner/ runner/
+COPY internal/ internal/
```
yitsushi, line 28: Why does the controller's Dockerfile need that directory. Not that it matters much.

internal/storage/storage.go
```diff
@@ -0,0 +1,115 @@
...
+	mw := io.MultiWriter(h, tf)
+
+	gw := gzip.NewWriter(mw)
+	tw := tar.NewWriter(gw)
+	if err := filepath.Walk(dir, func(p string, fi os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		// Ignore anything that is not a file (directories, symlinks)
```
yitsushi, line 55: I think symlinks can be used with Terraform, they can link modules, or just specific files between modules to make access "easier" to access and easier to maintain (edit only once).

internal/storage/storage.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		if _, err := io.Copy(tw, f); err != nil {
+			f.Close()
+			return err
+		}
+		return f.Close()
+	}); err != nil {
+		tw.Close()
+		gw.Close()
+		tf.Close()
```
yitsushi, line 93: I think we can put these `Close` calls into a defer, so we don't have to call `x.Close()` for all `if err := ..` block in the next few lines. Go doesn't cry about closing them multiple times.

```go
❯ cat main.go
package main

import (
	"compress/gzip"
	"crypto/sha1"
	"io"
	"log"
	"os"
)
❯ go run .


func main() {
	tf, err := os.CreateTemp("", "tf-")
	if err != nil {
		log.Fatal(err)
	}
	tmpName := tf.Name()
	defer func() {
		if err != nil {
			os.Remove(tmpName)
		}
	}()

	h := sha1.New()
	mw := io.MultiWriter(h, tf)
	gw := gzip.NewWriter(mw)

	gw.Close()
	gw.Close()
	gw.Close()
	gw.Close()
	gw.Close()
	log.Println("Done :)")
}
```

```
❯ go run .
2023/11/30 15:37:01 Done :)
```

internal/storage/storage.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if err := gw.Close(); err != nil {
+		tf.Close()
+		return "", err
+	}
+	if err := tf.Close(); err != nil {
+		return "", err
+	}
+
+	if err := os.Chmod(tmpName, 0644); err != nil {
```
yitsushi, line 110: - You can `Chmod` at the beginning, so the file has the desired permissions through the whole archiving process.
- I think `0600` is enough. Not that it has much value, but if we set permissions at all then why `0464`.

yitsushi (commented):
Partially reviewed, did not read tests yet.

yitsushi (approved):

PR #1162, opened 2023-12-07
feat: Implement BLOB encryption within the tf-runner

### LINKED ISSUE #1121: Implement BLOB encryption using service account tokens as keys within the tf-runner.

yitsushi:
As a starter, we need a ServiceAccountToken Secret. With older versions of Kubernetes, a default one was created automatically, but recent versions doesn't have this functionality.

```
apiVersion: v1
kind: Secret
metadata:
  name: tf-runner-encryption
  namespace: terraform
  annotations:
    kubernetes.io/service-account.name: tf-runner
type: kubernetes.io/service-account-token
```

When encrypting the BLOB content, we use the first 32 bytes of the generated token as AES256 key with a nonce using Galois Counter Mode.

The full encryption is part of the `CreateWorkspaceBlob` function for now. It can be extracted, but until we need somewhere else it can live there.

There is a chance we have to reconsider how it works when we approach the key rotation with #1161, because with this approach if the token is replaced, we have zero clue what was the previous one, therefore we can't decrypt existing content. It can be a design choice tho (discard all cache when key rotation happens).

Closes #1121

References:
- https://github.com/weaveworks/tf-controller/issues/1121
- https://github.com/weaveworks/tf-controller/issues/1161


runner/server_workspace_blob.go
```diff
@@ -33,14 +41,48 @@ func (r *TerraformRunnerServer) CreateWorkspaceBlob(ctx context.Context, req *Cr
 		return nil, err
 	}
 
+	secretName := "tf-runner.cache-encryption"
```
yiannistri, line 44: Shouldn't this match the secret name we're adding in the Helm templates of this PR i.e. `tf-runner-encryption`?

runner/server_workspace_blob.go
```diff
@@ -33,14 +41,48 @@ func (r *TerraformRunnerServer) CreateWorkspaceBlob(ctx context.Context, req *Cr
...
+		return nil, err
+	}
+
+	nonce := make([]byte, gcm.NonceSize())
+	_, err = rand.Read(nonce)
+	if err != nil {
+		return nil, err
+	}
+
+	out := gcm.Seal(nonce, nonce, blob, nil)
```
yiannistri, line 74: > // Seal encrypts and authenticates plaintext, authenticates the
	// additional data and appends the result to dst, returning the updated
	// slice. The nonce must be NonceSize() bytes long and unique for all
	// time, for a given key.
	//
	// To reuse plaintext's storage for the encrypted output, use plaintext[:0]
	// as dst. Otherwise, the remaining capacity of dst must not overlap plaintext.
	Seal(dst, nonce, plaintext, additionalData [][byte](https://pkg.go.dev/builtin#byte)) [][byte](https://pkg.go.dev/builtin#byte)

Why do we _need_ to append the result to the `nonce` (1st parameter)? Looking at the example in https://pkg.go.dev/crypto/cipher#example-NewGCM-Encrypt

yitsushi: Good question. I don't know :laughing:

yitsushi: Oh I know why, because that's how we save it so we can read it back as we don't have any other database or anything to save the nonce. It's there mostly to mess up dictionary attacks. It couldn't be a secret, but not necessarily a secret.

yitsushi: When decrypting we can read like this:
```
	nonce, ciphertext := resp.Blob[:nonceSize], resp.Blob[nonceSize:]
```

Without this, we would have to save the nonce somewhere else. Or use a static nonce value which feels a dumb thing to do, we had a secret key for encryption.

runner/server_workspace_blob.go
yitsushi: fixed in the chart, as I prefer `tf-runner.cache-encryption` over `tf-runner-encryption`, that describes much better the purpose of that key.

yitsushi: and follows the naming of the mtls secret:
```
terraform-runner.tls-1703067430   kubernetes.io/tls                     4      28h
```

runner/server_workspace_blob_test.go
```diff
@@ -2,38 +2,76 @@ package runner_test
...
+	g.Expect(err).To(BeNil())
+	blob, err := gcm.Open(nil, nonce, ciphertext, nil)
+
+	blobReader := bytes.NewReader(blob)
 
 	outputTempDir := t.TempDir()
-	untar.Untar(blobReader, outputTempDir)
+	sum, err := untar.Untar(blobReader, outputTempDir)
+	g.Expect(err).To(BeNil())
+	fmt.Println(sum)
```
yiannistri, line 74: We should probably change this to `t.Log` although in my local test run, this printed nothing/an empty line. Is that expected?

yitsushi: I think that should be empty, I'm not sure. But that package is already replaced by `tar`: https://github.com/fluxcd/pkg/tree/main/untar and `Untar` doesn't even return that value anymore. I think I used that value while I had issues with encryption/decryption and that contained value about the content of the byte stream.

yitsushi:
```
func untar.Untar(r io.Reader, dir string) (summary string, err error)

Untar reads the gzip-compressed tar file from r and writes it into dir.
```

Well, that's not so helpful. Maybe it was always empty and I never noticed it (and didn't remove the line) because it was empty :rofl:

runner/server_workspace_blob_test.go
```diff
@@ -2,38 +2,74 @@ package runner_test
...
+
+	// Decrypting content.
+	aesCipher, err := aes.NewCipher(key)
+	g.Expect(err).To(BeNil())
+	gcm, err := cipher.NewGCM(aesCipher)
+	g.Expect(err).To(BeNil())
+	nonceSize := gcm.NonceSize()
+	nonce, ciphertext := resp.Blob[:nonceSize], resp.Blob[nonceSize:]
+	g.Expect(err).To(BeNil())
+	blob, err := gcm.Open(nil, nonce, ciphertext, nil)
```
yiannistri, line 66: Should we swap these 2 lines? So that we check for errors after calling `gcm.Open`?

yiannistri (dismissed):
LGTM 👍

runner/server_workspace_blob_test.go
yitsushi: definitely

yiannistri (approved):
Looks great 🥇

PR #1168, opened 2023-12-22
Fix `make docker-build` target

yiannistri:
Fix `make docker-build` target as it was missing a TARGETARCH build argument. Note that this is only used in local dev, CI uses the official docker GitHub actions to build the images.
Also noticed that the env var for LIBCRYPTO_VERSION was misplaced.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,10 +12,10 @@ on:
 
 permissions:
   contents: read # for actions/checkout to fetch code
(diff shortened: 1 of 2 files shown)


yitsushi (dismissed):

yiannistri:
Actually my commit message was not entirely accurate. This is not only used for the terraform binary but also to build the controller/runner binaries for the alpine images. I'll reflect that in the Makefile and re-request a review. Sorry about the confusion.

yitsushi (approved):

PR #1170, opened 2023-12-27
fix use cases link in readme

nwilken:


yiannistri:
Thank you for fixing this :1st_place_medal:

yiannistri (approved):

Issue #338, opened 2022-09-06
Ability to provide multiple locations of .tfvars files with precedence


itamar-marom:
Managing multiple environments of one terraform module requires holding specific .tfvars (or more than one) for each environment.
We need the ability to provide the `Terraform` object, a list of paths from the same `sourceRef` that holds a `.tfvars` files and we also need to specify some kind of precedence between them. How terraform handles var-files:

```
"Any -var-file options on the command line, in the order they are provided. (This includes variables set by a Terraform Cloud workspace.)"
```

itamar-marom:
It can be a new field under Terraform's spec --> `varFiles`. It will hold a list of paths to `.tfvars` files and will add them in that order of precedence.

bivas:
If we're already on the subject, we should also support inline var input.

bivas:
These var files must be relative to the workspace and can be either pre exist in the cloned repository or mapped using the fileMapping option

itamar-marom:
After chatting with @bivas 

> If we're already on the subject, we should also support inline var input.

Already exists with the `Spec.vars` in Terraform object

There's a big question to talk about:
How do we see the precedence between `fileMappings`,` varsFrom`, `vars` and the new field we will create? BTW looks like there are too many ways to pass variables. Good because features are always welcomed, bad because it's confusing.

> These var files must be relative to the workspace

There are some cases where we want to use a tfvars file that is stored outside of the workspace but inside the `GitRepository` listed in `sourceRef`.

Therefore, We need to be able to specify paths that are relative to the `GitRepository` root directory.

options 1:
a list of paths and locations (similar to how we handle `fileMappings`), can be `workspace` or `root` (a better name for root can be `repository`, we need to decide). and the API will look like:
```
kind: Terraform
spec:
  tfvarsPaths: # Need to think of a good name.
    - location: "workspace"
      path: "./staging/override.tfvars" # relative to workspace
    - location: "root"
      path: ".meta/environments/staging/override.tf" # relative to git repository root
  sourceRef: # Already exists
    kind: GitRepository
    name: name
```

options 2:
we don't want to use `location`. That will require the controller to first try this path under `workspace` and if not exists will look under the git repository root. This options my cause some issues (same path in both workspace and root) and thus I don't support this one.
API will look like:
```
kind: Terraform
spec:
  tfvarsPaths: # Need to think of a good name.
    - "./staging/override.tfvars" # relative to workspace
[...]

PR #463, opened 2022-11-28
Feature/add tfvarsfiles spec

itamar-marom:
#338 
With this PR, we can add more `.tfvars` files to the Terraform object.
those files are only files located in the `Terraform.spec.sourceRef` and the path is relative to its root.
- Also, I upgraded the docker files base image to version 3.16.2 in order for it to build correctly.

Works the same as terraform's `-var-file`: https://developer.hashicorp.com/terraform/language/values/variables#variable-definition-precedence

Example:
```yaml
apiVersion: infra.contrib.fluxcd.io/v1alpha1
kind: Terraform
metadata:
  name: 
spec:
    path: deployments/eks/platform
    fileMappings: # File mappings will also be in the tfVarsFiles
        - location: "workspace"
           path: "another.tfvars"
           secretRef:
               name: ...
               key: ...
    tfVarsFiles: # This is the new field. It holds the precedence + paths
        - deployments/eks/platform/default.tfvars # From sourceRef
        - deployments/eks/platform/another.tfvars # From fileMappings
        - environments/platform/env.tfvars # From sourceRef
    sourceRef: # GitRepository as source
  	kind: GitRepository
        name: helloworld
        namespace: flux-system
```

PR #1185, opened 2024-01-24
Feature/add tfvarspaths spec2

itamar-marom:
New PR without conflicts.
Fix of PR #463 
Fix Issue #338 

Copy of old PR comment:

With this PR, we can add more `.tfvars` files to the Terraform object.
those files are only files located in the `Terraform.spec.sourceRef` and the path is relative to its root.
- Also, I upgraded the docker files base image to version 3.16.2 in order for it to build correctly.

Works the same as terraform's `-var-file`: https://developer.hashicorp.com/terraform/language/values/variables#variable-definition-precedence

Example:
```yaml
kind: Terraform
metadata:
  name: 
spec:
    path: deployments/eks/platform
    fileMappings: # File mappings will also be in the tfVarsFiles
        - location: "workspace"
           path: "another.tfvars"
           secretRef:
               name: ...
               key: ...
    tfVarsFiles: # This is the new field. It holds the precedence + paths
        - deployments/eks/platform/default.tfvars # From sourceRef
        - deployments/eks/platform/another.tfvars # From fileMappings
        - environments/platform/env.tfvars # From sourceRef
    sourceRef: # GitRepository as source
  	kind: GitRepository
        name: helloworld
        namespace: flux-system
```


yitsushi:
Thank you, I'll try to review in the following day. I'm a bit busy these days, but I'll try to finish as soon as possible.

runner/server_plan.go
```diff
@@ -134,6 +135,15 @@ func (r *TerraformRunnerServer) Plan(ctx context.Context, req *PlanRequest) (*Pl
 		planOpt = append(planOpt, tfexec.Target(target))
 	}
 
+	for _, path := range r.terraform.Spec.TfVarsFiles {
+		secureTfVarsFile, err := securejoin.SecureJoin(req.SourceRefRootDir, path)
```
yitsushi, line 139: Can we add a test where the referenced file doesn't exist.

There can be two routes if the file is not there:
1. We let terraform fail. If a file is not there, it should fail, but if terraform doesn't fail for some reasons, that can be an unexpected behaviour.
2. We check the file in this loop and return with an error if the file is not there. This one would be maybe better as we can give back a meaningful error message without other clutter from the command output.

In any cases, that would be good to see a test that fails intentional because of a missing referenced file.

What do you think?

itamar-marom: I agree. Will do.

itamar-marom: @yitsushi fixes + added a new test for non existing file

runner/server_plan.go
```diff
@@ -134,6 +135,28 @@ func (r *TerraformRunnerServer) Plan(ctx context.Context, req *PlanRequest) (*Pl
...
+		if err != nil {
+			log.Error(err, "Failed to secure join root dir with the given tfvars file's path.")
+			return nil, err
+		}
+
+		info, err := os.Stat(secureTfVarsFile)
+		if os.IsNotExist(err) {
+			log.Error(err, "The given tfvars file's path does not exists.")
```
yitsushi, line 147:
```suggestion
			log.Error(err, "The given tfvars file's path does not exist.")
```

runner/server_plan.go
```diff
@@ -134,6 +135,28 @@ func (r *TerraformRunnerServer) Plan(ctx context.Context, req *PlanRequest) (*Pl
...
+			return nil, err
+		}
+
+		if info.IsDir() {
+			log.Error(err, "The given tfvars file's path does not exists.")
```
yitsushi, line 152:
```suggestion
			log.Error(err, "The given tfvars file's path does not exist.")
```

runner/server_plan.go
```diff
@@ -134,6 +135,28 @@ func (r *TerraformRunnerServer) Plan(ctx context.Context, req *PlanRequest) (*Pl
...
+			return nil, fmt.Errorf("error running Plan: %s The given tfvars file's path is a directory",
```
yitsushi, line 153:
```suggestion
			return nil, fmt.Errorf("error running Plan: tfvars file's path is a directory: %s",
```

runner/server_plan.go
```diff
@@ -134,6 +135,28 @@ func (r *TerraformRunnerServer) Plan(ctx context.Context, req *PlanRequest) (*Pl
...
+		secureTfVarsFile, err := securejoin.SecureJoin(req.SourceRefRootDir, path)
+		if err != nil {
+			log.Error(err, "Failed to secure join root dir with the given tfvars file's path.")
+			return nil, err
+		}
+
+		info, err := os.Stat(secureTfVarsFile)
+		if os.IsNotExist(err) {
+			log.Error(err, "The given tfvars file's path does not exists.")
+			return nil, err
```
yitsushi, line 148:
```suggestion
			return nil, fmt.Errorf("error running plan: tfvars file's path does not exist: %s", secureTfVarsFile)
```

controllers/tc000360_tfvars_files_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+	It("should fail to plan.")
+	By("checking that the Plan's reason of the TF resource become `TFExecPlanFailed`.")
+	g.Eventually(func() map[string]interface{} {
+		err := k8sClient.Get(ctx, helloWorldTFKey, &createdHelloWorldTF)
+		if err != nil {
+			return nil
+		}
+		for _, c := range createdHelloWorldTF.Status.Conditions {
+			if strings.Contains(c.Message, "no such file") && strings.Contains(c.Message, "does-not-exists.tfvars") {
```
yitsushi, line 312:
```suggestion
			if strings.Contains(c.Message, "tfvars file's path does not exist: does-not-exists.tfvars") {
```

controllers/tc000360_tfvars_files_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+	It("should be reconciled and contain some status conditions.")
+	By("checking that the TF resource's status conditions has some elements.")
+	g.Eventually(func() int {
+		err := k8sClient.Get(ctx, helloWorldTFKey, &createdHelloWorldTF)
+		if err != nil {
+			return -1
+		}
+		return len(createdHelloWorldTF.Status.Conditions)
+	}, timeout, interval).ShouldNot(BeZero())
```
yitsushi, line 126: As we are not testing anything with these they are safeguards for the test run, we can merge them together, that saves a bit of time as it does not start a new `Eventually` loop. The second one would check the first one as well in the same loop.

controllers/tc000360_tfvars_files_test.go
```diff
@@ -0,0 +1,324 @@
...
```
yitsushi, line 302: Same as above, they can be merged.

yitsushi (changes requested):
Overall looks good to me. Some change suggestions (error message format, grammar), but looks good.

Thank you for the additional test case and checks for the requested file on the disk.

---

Errors should not contain capital letters unless that's a reference to an object (struct, interface, function) as they will be combined later with other error messages on the chain:

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.
>
> Source: https://go.dev/wiki/CodeReviewComments

controllers/tc000360_tfvars_files_test.go
yitsushi: It seems the two checks can't be merged as it can hit timeout, raising the timeout wouldn't be much better than having the two checks. Sorry about that, locally it was happy when I merged them, but I have a much stronger machine than CI has 😆

itamar-marom: Haha that's fine. I also remember wondering about this at the time my original PR but I didn't remember now why it was there. Now we both know 🙂
I'll add it back soon

yitsushi (approved):

itamar-marom:
@svengreb FYI
@yitsushi Thanks for reviewing this!

svengreb:
Thanks both of you, for implementing this feature and picking the review back up so quickly 🚀

moiferrsan:
Hi guys,

Is this feature already available?

It will be very useful

yitsushi:
It's available on main, but we don't have a release yet that contains this patch.

moiferrsan:
Thanks @yitsushi

PR #1197, opened 2024-02-26
replace container reg ghcr.io/weaveworks with ghcr.io/flux-iac

dgem:
Assumes that the container reg is setup for the new org


dgem:
sorry, wrong issue number of the branch name... will rename

Issue #1200, opened 2024-02-26
change image references to reference the new organization


ilithanos:
When the build pipeline fix is merged, the next logical step would be to update the toolings and deployments to pull images from this organization instead of the weaveworks organization.

Current places i've found so far:

- [ ] Helm Chart
- [ ] Tilt
- [ ] Makefile
- [ ] tools/check-runner-images.sh
- [ ] config/*
- [ ] controllers/tf_controller_runner.go
- [ ] controllers/tc000260_runner_pod_test.go
- [ ] Lots of places in documentation

ilithanos:
i'll start going through this list and fixing what i find.

Will do it in multiple smaller PR's so reviewing and testing is easier.

ilithanos:
@chanwit When going over the helm chart how much rebranding to tofu controller is wanted?

For now my thought is fixing the gcr urls and the helm repo urls and leaving it at that. But if you want every tf-controller string/reference to be changed to tofu-controller i'll do that while i'm at it.

chanwit:
Hi @ilithanos 

agreed
the first priority is to make things work, so it's fine to fix only things needed.

ilithanos:
I haven't gone through the documentation again yet, to evaluate if we are some with that part.

Tilt is still missing a few things. I have no experience in using tilt, but I'll see what I can come up with the following days.

chanwit:
Tilt was very useful to us when debugging a complex controller + runner behaviour.

ilithanos:
Good to know, I'll have to try it out. It's just one of the tools I haven't run into yet.

@kasper-leFevre could I get you to verify that the documentation references are all fixed?

@akselleirv I've noticed you are digging into the Tilt setup, would you mind taking on the registry references etc while you are at it?

Just trying to get an overview of what we still need to get done on this one so we can get it finished up. Thanks to everyone for helping out ❤️

akselleirv:
@ilithanos I made the Tilt setup functionally work in https://github.com/flux-iac/tofu-controller/pull/1224/files.

I can see that there are some references to tf-controller. I can change them to tofu-controller.

chanwit:
ping @ilithanos 
could we have a quick update of this issue please.
what are left to do to make it complete?

cc other maintainers @akselleirv @yitsushi

akselleirv:
Hello @chanwit, I’m at KubeCon this week so a bit busy, but I’ll take a look at this next week if that is fine.

[...]

PR #1213, opened 2024-02-28
move urls from weave to flux-iac in makefile

ilithanos:
This fixes Makefile part of #1200.

I havn't tested this fully yet, so making this as a draft PR for now as it might have unforseen side effects.

will do some testing during today and see what happens.


ilithanos:
Changing it to ready for review since all the pipelines that uses the makefile still seem to work.

chanwit (approved):
LGTM!

PR #1214, opened 2024-02-28
updating documentation for custom runner

kasper-leFevre:
updated following in the documentation:

- link to base image in ghcr (not yet available but ready for when it is)
- tf version to reflect the tf version there is used
- changed link to original dockerfile from a specific commit to main branch


chanwit (approved):
Thank you @kasper-leFevre !

LGTM

PR #1216, opened 2024-02-28
config yaml renaming and docker file binary renamed

### LINKED ISSUE #1215: Replace tf-controller to tofu-controller in manage.yaml
We now have to start touching the manage.yaml for example:

https://github.com/flux-iac/tofu-controller/blob/b2185903ae4230559b0a2f2d8bd29de6cb2e398b/config/manager/manager.yaml#L26

every `tf-` reference should be change to `tofu-` except `tf-runner`.

maybe we need to be careful about these values in Helm charts too.

ilithanos:
Rebranding/renaming to tofu-controller and flux-iac in /config

I've purposefully left service accounts alone, due to the way they are branded with kustomize resulting in also renaming the tf-runner accounts.

updated binary name in the docker file to tofu-controller.

Should fix #1215 and part of #1200

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -58,16 +58,16 @@ jobs:
             echo 'run make test and commit changes'
             exit 1
           fi
-      - name: Build the tf-controller container image
+      - name: Build the tofu-controller container image
         run: |
           VERSION="e2e-${GITHUB_SHA::8}"
           docker buildx use default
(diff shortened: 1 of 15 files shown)


ilithanos:
Looking into fixing the e2e test, seems like some name references are hardcoded into the tests.

ilithanos:
namespaces and tf-controller naming was hardcoded into the e2e tests and makefile. 
i've updated both to the tofu naming.

chanwit (approved):
Thank you so much @ilithanos 

LGTM

PR #1220, opened 2024-02-29
dockerfiles updated to alpine:3.19

### LINKED ISSUE #1219: Update Alpine in base image to make it pass the CVE scanner
We currently use a quite outdated version of Alpine. That's why we got many CVEs in OS packages. 

https://github.com/flux-iac/tofu-controller/blob/main/runner-base.Dockerfile#L41

The version of Alpine needs to be updated so that we get new enough OS packages to avoid CVEs.

ilithanos:
Fixes #1219 

There is a but attached to this though.

updating to alpine 3.19 seems to have updated packages to a python version that follows [https://peps.python.org/pep-0668/](pep-668) which enforces python virtual environments.

The current implemented solution is to add the --break-system-packages flag to the commands. This basically bypasses pep-668 and locale2e tests passes, but it might be good to consider moving it to virtual environments instead.

### DIFF
--- a/.github/workflows/scan.yaml
+++ b/.github/workflows/scan.yaml
@@ -46,13 +46,13 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
-      - name: Build tf-controller image
+      - name: Build tofu-controller image
         run: |
           make docker-buildx
       - name: Run Trivy vulnerability scanner on controller image
         uses: aquasecurity/trivy-action@fbd16365eb88e12433951383f5e99bd901fc618f # v0.12.0
         with:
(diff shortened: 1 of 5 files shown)


ilithanos:
Am i reading the pipeline wrong or is it building the image, but pulling the old one from the ghcr when running the Trivy test? 

As far as i can tell it will always scan the latest image pushed to the registry rather than the one related to the PR?

chanwit:
@ilithanos you're right.
the pipeline built `ghcr.io/flux-iac/tofu-controller:latest` but scanned `ghcr.io/flux-iac/tf-controller:latest`

ilithanos:
There we go, all green now :)

chanwit:
woot!! :tada: :tada: :tada: :tada: :tada:

chanwit (approved):
LGTM

PR #1221, opened 2024-02-29
#1209 Relocate CRDs in helm chart

### LINKED ISSUE #1209: Better CRDs handling
I had an issue while upgrading the TF controller this morning. It’s deployed using **Terraform**.

The current way tf-controller helm chart is using to install CRDs is using a template file, [located](https://github.com/flux-iac/tofu-controller/blob/a56c128f070eff1d8252b36fdb1590f34ffad298/charts/tf-controller/templates/crds.yaml) in the `templates` folder. This can cause issue when uninstalling the chart: the CRDs gets deleted, which in turn can trigger a "terraform destroy" on the corresponding CR. I have observed this behavior upgrading from tf-controller to tofu-controller. My Terraform CR got deleted and underlying resources destroyed. The CRD status turned into "Terminating", waiting for the finalizer to finish destroying resources, as seen in this k9s screenshot (note I rolled back to previous tf-controller version before the end of CR termination)

image

According to Helm [documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations), the standard way to deploy CRDs on a cluster is to locate them in a `crds` folder. By design, `helm uninstall ...` won't touch the CRDs defined there, so it's safer to upgrade the controller in case, for whatever reason, it's required to uninstall and reinstall it. In current deployment, CRDs are just uninstalled from k8s, just like any other "normal" objects.

dgem:
Per #1209 , this removes the use of templates in the crds and moves them to `./charts/tf-controller/crds`

Need to decide whether to replicate lables that were previously generated using the helper template as hard-coded values. In this PR there are two included as examples:
```
  labels:
    helm.sh/chart: tf-controller-0.16.0-rc.3
    app.kubernetes.io/version: 0.16.0-rc.3
```

This is a 'fix forward', it won't fix the issue that uninstalling previous versions of this or its predecessor can trigger a destroy.

Close #1209


chanwit:
```
[FATAL tini (13)] exec tf-controller failed: No such file or directory
```

related logs above

dgem:
@ilithanos, assuming I'm reading your comment correctly, I can see in the [deployment.yaml#L49](https://github.com/flux-iac/tofu-controller/blob/main/charts/tf-controller/templates/deployment.yaml#L49):
```yaml
        command:
        - /sbin/tini
        - --
        - tf-controller
 ```
 I've updated this...
 
With regards to image names, is that a case of syncing the names in the `values.yaml` with values in the top of the makefile:
```
MANAGER_IMG ?= ghcr.io/flux-iac/tofu-controller
RUNNER_IMG  ?= ghcr.io/flux-iac/tf-runner
RUNNER_AZURE_IMAGE ?= ghcr.io/flux-iac/tf-runner-azure
BRANCH_PLANNER_IMAGE ?= ghcr.io/flux-iac/branch-planner
```
?

ilithanos:
when updating the helm docs, run make helm-docs rather than manually editing the .md file, as it's generated based on the go.tpl file. The check will fail unless it's exactly the same as the generated file would be.

ilithanos:
@chanwit LGTM now, the Trivy scan is unrelated.

chanwit (approved):
LGTM Thank you @dgem 
and @ilithanos for the review!

PR #1223, opened 2024-03-01
fallback ghcr urls changed to flux-iac in runner

ilithanos:
Changes the default url for the runner images to point to the flux-iac registry in the go code and test.

This fixes 2 tasks in #1200


chanwit (approved):
LGTM

Thank you @ilithanos

PR #1224, opened 2024-03-04
Tiltfile: added config for building tf-runner image and using it in helloworld example

akselleirv:
The tf-runner image was only built but not used in the helloworld-tf Terraform example. With this change the Terraform spec will be updated by Tilt with the tag which was produced by `docker_build`.


chanwit:
Thank you @akselleirv 
Let me know if it's ready to review.

akselleirv:
@chanwit it is ready to be reviewed

chanwit (approved):
LGTM 

Thank you @akselleirv :1st_place_medal:

PR #1226, opened 2024-03-05
CLI: Reduce Binary Size

bdwyertech:
This adds a Windows build for the CLI, and also reduces the resulting binary size by removing debug symbols.  The yield is about 30% reduction in size.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -15,8 +15,11 @@ builds:
     binary: tfctl
     env:
       - CGO_ENABLED=0
+    flags:
+      - -trimpath
     ldflags:
-      -  -X main.BuildSHA={{.ShortCommit}} -X main.BuildVersion={{.Tag}}
+      - -s -w
+      - -X main.BuildSHA={{.ShortCommit}} -X main.BuildVersion={{.Tag}}
     goos:
       - darwin
       - linux


chanwit:
@ilithanos @yitsushi what do you think?

yitsushi:
Removing the binary size is a good addition, but adding windows build can cause some issues.

We have no mechanisms to test windows builds, and we (at least I don't) have a place to test it manually to make sure it's working properly as a result I can't say we can support it. If it doesn't work we can't even debug it. Adding a windows build would indicate we support it, but we have no capacity and resources to support it. At least that's my opinion. It can be a good addition, but is no one can guarantee we can support it that may result removing it later if it's broken at any points in time. If we add docs around it like "the windows build is not 100% supported, if something is broken with it, we may not be able to debug and fix it", but that would give a bad taste too.

ilithanos:
The binary size reduction is good.

Regarding the windows build i completely agree with your assessment. We don't have a way to debug it, test it or maintain it. The idea is good, but at the same time i have a hard time seeing the need for it. Since the linux build could be run on a windows machine using WSL without any changes.

Having released versions that we can't support seems like a complete no go for me. I would rather document how to run the linux binary in WSL on windows, or throw a docker container on Windows considering the current state of the project, and the current amount of resources available for maintaining it.

bdwyertech:
Understood.  I'll keep the Windows binaries on my fork.  Macs are a rarity for devs around here unless you are special (large enterprise).

ilithanos:
I have seen windows used by developers in some cases, but mostly in Azure/DotNet/c# based companies. And in those cases they always had WSL, Docker or a Linux VM in use as well.

At the current place only C level management, support, sales etc are running windows, everybody in the development related departments are offered the choice between Mac or Linux machines ;) 

@bdwyertech is there a reason a Docker based or WSL based solution isn't an option? If for nothing else then just to expand my horison :)

bdwyertech:
> I have seen windows used by developers in some cases, but mostly in Azure/DotNet/c# based companies. And in those cases they always had WSL, Docker or a Linux VM in use as well.
> 
> At the current place only C level management, support, sales etc are running windows, everybody in the development related departments are offered the choice between Mac or Linux machines ;)
> 
> @bdwyertech is there a reason a Docker based or WSL based solution isn't an option? If for nothing else then just to expand my horison :)

Proxies, VPNs etc...  I own the CLI tool internally and I've had to add a `doctor` function to check registry entries, proxy configs, env vars, docker configs, etc. to make sure everything looks good.

15k employee enterprise, I think I am one of 10 dev's with a Mac, about 100 macs total - the rest are executives.

ilithanos:
Oh, yeah i can see that being an issue. I would say exposing tfctl to developers without understanding of basic k8s etc might be a risk :)

If using github branch planner might also be an option. I remember they days of using AnyConnect with dread. It works, but is very finicky about the environment it runs in :) 

Would you mind either cleaning up the PR by removing the windows specific build, but leaving the binary size optimization? then i'm sure @chanwit will merge it when he gets around to it.

Thanks for helping out with the project and giving me some more info on the landscape of the userbase.

chanwit:
@yitsushi is also a maintainer who could also help review this issue in my absence.

bdwyertech:
> Oh, yeah i can see that being an issue. I would say exposing tfctl to developers without understanding of basic k8s etc might be a risk :)
> 
> If using github branch planner might also be an option. I remember they days of using AnyConnect with dread. It works, but is very finicky about the environment it runs in :)
> 
> Would you mind either cleaning up the PR by removing the windows specific build, but leaving the binary size optimization? then i'm sure @chanwit will merge it when he gets around to it.
> 
> Thanks for helping out with the project and giving me some more info on the landscape of the userbase.

Sure, no prob.  I’ll clean it up later this afternoon.

100% agree with the risk aspect.  This tool feels more geared to a platform engineer than something Joe dev needs.  
I’m exploring possibilities here more than anything else.  This project looks great and I love the execution model where per-project pods get stood up to scale tf execution horizontally (and offering unique permissions by way of IRSA annotation).  I’ve contributed a few features to the tf crossplane provider, quick glance says this project is more in line with our plans.  Should this be a path we take, you’ve got another contributor for sure.

bdwyertech:
Done.  Didn't rename the branch to avoid opening a new PR and losing this convo.

yitsushi (approved):

chanwit (approved):
LGTM

Thank you @bdwyertech !!

PR #1239, opened 2024-03-13
helm-chart: add watchAllNamespaces argument

akselleirv:
Make it possible to set the `--watch-all-namespaces` flag to false


chanwit (approved):
LGTM

Thank you @akselleirv !!

PR #1248, opened 2024-03-14
Update development docs

akselleirv:
Closes https://github.com/flux-iac/tofu-controller/issues/1236

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -312,3 +312,12 @@ docker:
 .PHONY: serve-docs
 serve-docs: ## Run a local server to serve the docs
 	@docker run --rm -it -p 8000:8000 -v $(shell pwd):/docs squidfunk/mkdocs-material
+
+.PHONY: up
+up:
+	./tools/reboot.sh
+	tilt up
+
+down:
+	kind delete cluster --name tfdev
+	docker rm -f kind-registry
--- a/Makefile
+++ b/Makefile
@@ -312,12 +312,3 @@ docker:
 .PHONY: serve-docs
 serve-docs: ## Run a local server to serve the docs
 	@docker run --rm -it -p 8000:8000 -v $(shell pwd):/docs squidfunk/mkdocs-material
-
-.PHONY: up
-up:
-	./tools/reboot.sh
-	tilt up
-
-down:
-	kind delete cluster --name tfdev
-	docker rm -f kind-registry
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -70,43 +70,31 @@ to update the the generated `runner.pg.go` data.
 
 ## How to run the controller locally
 
-Install flux on your test cluster:
+Prerequisites:
+
+* go
+* kind
+* kubectl
+* tilt
+* docker
+* flux
+
+Start the KinD cluster with a local registry and install flux:
 
 ```bash
-flux install
+./tools/reboot.sh
 ```
 
-Port forward to source-controller artifacts server:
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -312,3 +312,12 @@ docker:
 .PHONY: serve-docs
 serve-docs: ## Run a local server to serve the docs
 	@docker run --rm -it -p 8000:8000 -v $(shell pwd):/docs squidfunk/mkdocs-material
+
+.PHONY: up
+up:
+	./tools/reboot.sh
+	tilt up
```
akselleirv, line 319: Need some input how to best handle this.

The Tilt command will fail here as it is missing the GITHUB_TOKEN required by the branch planner.
Should it fail if it is not set, or let the let the tilt up command fail?

Could also just call the reboot.sh and let the user type in `tilt up` after calling make up.

chanwit: @yitsushi what do you think?

yitsushi: I wouldn't combine `reboot` script and `tilt up` in this way as I think it's misleading. That script wipes out the whole kind cluster and creates a new one. A `make` target called `up` does not suggest that. I know it's much longer, but `reboot-and-up` (any anything else that indicates the reboot).

For failing or not. It sounds reasonable to fail, and that's another reason why it's kind of useless to combine the two commands as one target. If `tilt up` fails hard, you may have to issue `tilt up` manually again, but you don't want to recreate the whole kind cluster, that sounds like a huge waste of time and resources. With this the user may use `make up`, if it fails up in shell history and execute `make up` again, but in reality following the docs they would create a kind cluster and just use `tilt up`.

Ideally you run `tilt up` whenever you stopped the command execution (`^C`) for example to force refresh all resources, changed something fundamental in the the Tiltfile. None of these cases involves recreating the whole kind cluster. For comparison:
1. I don't think I used reboot that much, docs around it seems more reasonable. I used it like 2-3 times in 3 months, but definitely not more than 5 times.
2. I used `tilt up` like 2-3 times per day on average when worked on it full time.
3. If something was so messed up, usually a `tilt down` and `tilt up` was more than enough to clean up the environment.
4. You can terminate the `tilt up`, the cluster is still alive, but Tilt will not update resources, which can be a huge resource saver if you plan to change a lot of things at once, you can kill the `tilt up` process, make your changes for the next X minutes or even hours, without the cluster getting updated all the time in a state you are certain it will fail to update or even compile. When you are done, you can start `tilt up` again and be happy. This way you saved a lot of resources not forcing a tilt resource update whenever you saved a Go file with changes that may even render the code unusable (unable to compile) as your are still in the middle of a bigger refactor / feature implementation.

yitsushi: Now to mention, kind is a totally optional dependency, it can be anything that can run a cluster, I know some folks are using microk8s, or having a full testing k8s cluster somewhere (don't ask why, but possible), the workflow has zero changes in that case. Kind is there to create a cluster in an easy way, and tilt doesn't care about it at all, it'll just use the active k8s context (or one defined in an Env variable).

akselleirv: Does it make sense to remove these commands from the Makefile and just document them in the DEVELOPMENT.md file?

It's not too bad as there are only two commands to remember.

yitsushi: Yes.

akselleirv: Removed it now and its ready for review @chanwit

chanwit (approved):
LGTM

Thank you @akselleirv !

PR #1251, opened 2024-03-14
Fix runner-serviceaccount helm template

ayanevbg:
The end statement for the custom imagePullSecrets is wrongly placed af the bottom of the yaml file. Instead it should be placed at the end of the ServiceAccount definition.


chanwit (commented):
LGTM

Thank you so much @ayanevbg :+100:

chanwit (approved):

PR #1259, opened 2024-03-17
chainsaw: added poc test

akselleirv:
Relates to https://github.com/flux-iac/tofu-controller/issues/1131

This is a rewrite of the main parts of the [controllers/tc000010_no_outputs_test.go](https://github.com/flux-iac/tofu-controller/blob/main/controllers/tc000010_no_outputs_test.go) test to showcase the benefits of [Chainsaw](https://github.com/kyverno/chainsaw).

Resources can be reused across the tests by placing the YAML manifests in a shared folder, instead of inlining them like I have done here. I find the declarative approach to testing much more readable compared to the more imperative Go test.

The current test uses `oras` to pack and push the OCI image to the local kind registry, but can be replaced with for example the `flux` CLI if we would like to avoid the dependency.
 
The test can be run with the following commands:

```bash
./tools/reboot.sh 
tilt up
chainsaw test test/no-outputs/
```

Is this approach something that could replace parts of the current controller test suite? Or would it be only be suitable for replacing the e2e tests?

### DIFF
--- a/test/data/terraform/hello-world-example/main.tf
+++ b/test/data/terraform/hello-world-example/main.tf
@@ -1,0 +1,11 @@
+terraform {
+  # This module is now only being tested with Terraform 0.13.x. However, to make upgrading easier, we are setting
+  # 0.12.26 as the minimum version, as that version added support for required_providers with source URLs, making it
+  # forwards compatible with 0.13.x code.
+  required_version = ">= 0.12.26"
+}
+
+# website::tag::1:: The simplest possible Terraform module: it just outputs "Hello, World!"
+output "hello_world" {
+  value = "Hello, World!"
+}
--- a/test/no-outputs/chainsaw-test.yaml
+++ b/test/no-outputs/chainsaw-test.yaml
@@ -1,0 +1,88 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/kyverno/chainsaw/main/.schemas/json/test-chainsaw-v1alpha1.json
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: no-outputs
+spec:
+  namespace: terraform
+  description: "This spec describes the behaviour of a Terraform resource with no backend, and `auto` approve."
+  steps:
+    - name: setup source
+      try:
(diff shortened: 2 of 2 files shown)


chanwit:
The test structure is so clean and very easy to understand. Thank you so much @akselleirv 

What do you think of putting a test number as the prefix of dir no-output?

We found that test case numbering gave us a good way to organizing, grouping while keeping the flat structure of test cases at the same time.

chanwit:
Since @yitsushi spent a load of time restructuring tests in CI, I'd to hear your options too.

akselleirv:
> What do you think of putting a test number as the prefix of dir no-output?

I can definitely do that! That makes it easier to run them in parallel groups like you have done for the existing tests.

yitsushi:
I'm not against chainsaw itself, but in my opinion it should not replace go tests.

For e2e is a different question that's totally fine, but `controllers/tc*` are not e2e, they are (supposed to be) unit tests.

---
A totally opinion based, 100% personal take on the topic:
Speaking of myself, I would not write tests in this format. Nothing special, it's just not how I would work on any projects and if a project would require me to write yaml files as test, that would discourage me to even think about contributing to that project. I'm well aware it's a personal issue, but I can't help that. I like when I have the least amount of abstraction on top of the tools I'm using. As I write Go code, I expect Go tests, a test code that can be changed easily if required. Abstraction layers make it much harder to change things, they force the user in a give way and "deal with it" if that doesn't fit. I'm totally find adding extra tests, but I prefer to have go tests for everything that's written in go. To remove the ugliness of inline yaml files, there are other options (for example using fixtures directory, a module with specific structure, using `embed`, etc).

akselleirv:
I definitely see your point @yitsushi!  

Maybe start by replacing the current e2e test workflow to chainsaw. I think it’s easier to read and then the e2e test can be executed locally instead of having to push it to GitHub in order to trigger the workflow.

Then later on extract the more e2e go tests within the `controllers/` to be using Chainsaw?

akselleirv:
I'll close this PR and then we can consider using Chainsaw for e2e tests in the future. Thanks for all the feedback!

Issue #1252, opened 2024-03-14
Renaming Helm chart to `tofu-controller`


chanwit:
Please let me know if it's going to break your current setup.

Sierra1011:
@chanwit Apologies for nagging, it's not meant that way; I saw that #1264 merged last week but the Action to release the updated HelmChart versions/name hasn't run. Is there a specific hold due to external events (_cough xz_) or is this just imminently in the pipeline? I'm keen to update :grin:

ilithanos:
@Sierra1011 the helm charts follow the tofu-controller versioning tightly, so it will be released together with the next rc version.

There are a few more PR's that I would like to have merged into that rc version before releasing, but it will be done soon ;)

akselleirv:
@ilithanos do you have any more changes before we can create a new release?

PR #1264, opened 2024-03-24
Helm chart rename

### LINKED ISSUE #1268: tfVarsFiles is missing in helm chart crds
Hi,

I was trying to use tfVarsFiles feature using helm chart 0.16.0-rc.4, during flux reconcile I get error that tfVarsFiles is missing in schema. I further investigate the issue and find out that tfVarsFiles crd specification:
https://github.com/flux-iac/tofu-controller/blob/main/config/crd/bases/infra.contrib.fluxcd.io_terraforms.yaml#L9815

is missing in helm chart crd (and also the version in this file is not set to latest 0.16.0-rc.4):
https://github.com/flux-iac/tofu-controller/blob/main/charts/tf-controller/crds/crds.yaml

ilithanos:
#1252 has no comments on this.

This should rename the helm chart to the new tofu-controller naming and ensure versions in values and labels are tagged with the rc.4 version of the tofu controller.

It should also change the CI to use the new naming convention for both chart and controller

Also fixes #1268 by updating the CRDs within the chart


ilithanos:
Not fully ready for review yet. Will add reviewers when all tests etc are working.

ilithanos:
Hmm interesting that this change breaks all the tests with a go version syntax error. I'll look into what is going on.

ilithanos:
The tests starting to fail wasn't related to this specific branch, but would happen in all branches:

https://github.com/kubernetes-sigs/controller-runtime/issues/2720

Using the latest tag for pulling in versions in CI pipelines are in general bad practise. I've pinned the version of setup-envtest to solve the issue for now. 

Will create a seperate issue for this topic instead of going into to much detail here.

ilithanos:
While renaming the chart i also updated the CRDs with the changes from config/crd/bases which should fix #1268

ilithanos:
This should be ready for review, but i might have missed something that could have side effects without knowing. 

Especially when it comes to the CRDs. So any feedback would be appreciated.

chanwit (approved):
LGTM

PR #1272, opened 2024-04-03
Tilt: fix references to helm chart after renaming

akselleirv:
Fix references to the old helm chart named `tf-controller` to the new name `tofu-controller`.


ilithanos (approved):
Looks good to me :)

PR #1273, opened 2024-04-03
feat: added support for tofu

akselleirv:
Adds initial support for tofu by making the `execPath` setting configurable. It can be configured globally or per Terraform resource.

I also had to bump the controller-gen binary as I was not able to run `make manifests` with the old version, which resulted in some shifting of the descriptions in the CRD.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -247,7 +247,7 @@ protoc-gen-go-grpc: ## Download controller-gen locally if necessary.
 CONTROLLER_GEN = $(GOBIN)/controller-gen
 .PHONY: controller-gen
 controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.9.2)
+	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.14.0)
 
 # Find or download gen-crd-api-reference-docs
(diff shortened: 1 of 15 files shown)


ilithanos:
Only thing I can see blocking this is that trivy security warning related to the tofu binary. 

I'm unsure though if this is the way we want to do it, as another option is using the tofu binary and tofu-exec package instead of tf-exec as well in the runner.

My understanding (@chanwit correct me if I'm wrong) is to change to OpenTofu as the native solution with secondary support for hashicorp Terraform. And if that is the case a move to tofu-exec and tofu binary would be the way to do that.

We could choose to support both with different runner images as well, as there's no need to include both binaries in the same image. 

I'm not opposed to this PR's solution, but I think we should have a discussion about how to go about this in the long run.

akselleirv:
Apologies for my late response @ilithanos !

> I'm unsure though if this is the way we want to do it, as another option is using the tofu binary and tofu-exec package instead of tf-exec as well in the runner.
> My understanding (@chanwit correct me if I'm wrong) is to change to OpenTofu as the native solution with secondary support for hashicorp Terraform. And if that is the case a move to tofu-exec and tofu binary would be the way to do that.

I might have been a bit too quick with creating a PR for adding support for OpenTofu without discussing the best approach with you guys. Do we have an issue where we can continue the discussion of how to implement OpenTofu? Then I close this PR and we can come back to it after we have a consensuses regarding the implementation.

ilithanos:
Seems like this kind of discussion is normally done using the GitHub discussions on this project and not within issues, so let's create one and take it from there ;)

Issue #490, opened 2023-01-05
Docs Issues


dfreilich:
I've had a few issues with the TF Controller in the past few days, and wanted to give some suggestions to the docs as a result: 
* We set `autoApply` to the only other enum value, which was `disable`, and didn't realize that that would disable planning as a result (as documented [here](https://github.com/weaveworks/tf-controller/blob/a4d31bb311e7d405bb9e0d3b357b4bb7766927b5/docs/use_tf_controller/to_detect_drifts_only_without_plan_or_apply.md)). It would've been helpful if that was also documented [here](https://github.com/weaveworks/tf-controller/blob/a4d31bb311e7d405bb9e0d3b357b4bb7766927b5/docs/use_tf_controller/to_plan_and_manually_apply_Terraform_resources.md#use-tf-controller-to-plan-and-manually-apply-terraform-resources), so the use of that flag would be clear.
* The documentation about custom-backends [here](https://github.com/weaveworks/tf-controller/blob/a4d31bb311e7d405bb9e0d3b357b4bb7766927b5/docs/use_tf_controller/with_a_custom_backend.md) doesn't mention the ability to add fields from secrets or configMaps, as mentioned in this [issue](https://github.com/weaveworks/tf-controller/issues/290)
* There's little documentation on what certain fields in the spec do, like `force`. Even just adding more docs in the code comments, so that they show up in the API Documentation, could be really helpful (for instance, before we found [these docs](https://github.com/weaveworks/tf-controller/blob/a4d31bb311e7d405bb9e0d3b357b4bb7766927b5/docs/use_tf_controller/to_force_unlock_Terraform_states.md) on `forceUnlock`, we went hunting through the code to understand the behavior)

Thanks for all the work on it!

chanwit:
Thank you so much @dfreilich !!

I'll improve the docs per your suggestions.

LappleApple:
Hello @dfreilich, it's been a while. :) Would you be open to making a PR to implement this change? Thanks much.

PR #1275, opened 2024-04-06
Improve docs

vishu42:
#490

### DIFF
--- a/docs/use-tf-controller/with-a-custom-backend.md
+++ b/docs/use-tf-controller/with-a-custom-backend.md
@@ -39,10 +39,56 @@ spec:
       image: registry.io/tf-runner:xyz
 ```
 
-To add fields from secrets or configMaps, use `-backend-config` arguments of the `init` command to set the values:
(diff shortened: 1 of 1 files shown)


vishu42:
@chanwit I have opened this PR to address issue #490

- I don't think any change is required for 1st point, as the doc [here](https://github.com/flux-iac/tofu-controller/blob/main/docs/use-tf-controller/detect-drifts-only-without-plan-or-apply.md) is pretty clear on the behaviour of `.spec.approvePlan` :`disable`. 
- For the 2nd point I have made the required changes in the document.
- For the 3rd point, the scope seems very broad as it questions documentation of each field in terraform spec. Also, the [spec](https://github.com/flux-iac/tofu-controller/blob/main/docs/References/terraform.md) has been updated many times since the issue was originally reported which means many of the reporter's concerns might have already been resolved.

vishu42:
@chanwit can you please take a look :)

chanwit (approved):
yep, LGTM.
Thank you @vishu42

PR #1282, opened 2024-04-16
fix(helm-chart): missing renamings

akselleirv:
Closes https://github.com/flux-iac/tofu-controller/issues/1280


akselleirv:
Seems to be some unrelated issue of building the `runner-azure.Dockerfile`.

ilithanos (approved):

ilithanos:
Seems like all the failing pipelines are due to the same issue with building the container images.

Looks like a package conflict where some packages will break the openssl package used.

I've approved the helm chart changes, let's fix the build issues in a separate PR ;)

Issue #1068, opened 2023-10-26
Incorrect gotk_reconcile_condition metric after migration to v0.16.0-rc.3


renato0307:
Hi,

I've upgraded from `v0.14.0` to `v0.16.0-rc.3`.

The `gotk_reconcile_condition` for the `Ready` condition is not being correctly reported.
* Terraforms are Ready
* Metric reports Unknown=1 and True/False=0

Do you have any idea the cause? Could you please help me troubleshoot this issue?

Thank you in advance.

More details:

### v0.16.0-rc.3

![image](https://github.com/weaveworks/tf-controller/assets/2320628/5223a555-96d9-45d0-817a-8357fb817970)

![image](https://github.com/weaveworks/tf-controller/assets/2320628/4de00cee-62d8-49ac-a3ec-a4d09dec9fbf)

![image](https://github.com/weaveworks/tf-controller/assets/2320628/2bb6ed6d-0cca-4a02-9efd-d62bf64b375f)

### v0.14.0

![image](https://github.com/weaveworks/tf-controller/assets/2320628/cba2823e-2d7d-4a33-ac95-6dfb095c57f4)

![image](https://github.com/weaveworks/tf-controller/assets/2320628/4a1f812d-9d7f-4f49-9232-c791577cebff)

renato0307:
Ran the controller with `log-level=trace`:

![image](https://github.com/weaveworks/tf-controller/assets/2320628/317fb166-3611-406c-94f1-94e5deee3f74)

chanwit:
Thank you @renato0307 for reporting this.

Maybe it's because we have recently changed the code base to match that of Flux v2 GA.

@luizbafilho @yitsushi could you help take a look please?

cc @LappleApple

renato0307:
Anyone working on this @LappleApple @chanwit? May I give it a go?

LappleApple:
Hey @renato0307, all yours. :) I'll assign it to you.

yitsushi:
@renato0307 : Go for it, if any of us picked it up, we assign ourselves.

renato0307:
Looks like the problem is on the [defer](https://github.com/weaveworks/tf-controller/blob/1958307fc83cc56f486453885e9426cd88f93d80/controllers/tf_controller.go#L133).

[...]

PR #1287, opened 2024-04-26
Fix Incorrect Metric Reporting Post-Update in Terraform Controller

TarasLykhenko:
Fixes: #1068

This PR addresses a bug in the Tofu Controller where the gotk_reconcile_condition metrics for the Ready condition were not being accurately reported, causing the metrics to perpetually show as updating even after a successful update.

Issue:
Post-update, the metrics incorrectly reported Unknown=1 with True/False=0, due to the misuse of the original terraform variable in a deferred function. This variable held the Unknown condition and inadvertently overwrote the correctly updated metrics.

Root Cause:
The defer statement used the original terraform variable, which contained the Unknown condition. This action overwrote the metrics that were updated correctly elsewhere in the code at a different point ([see here](https://github.com/weaveworks/tf-controller/blob/1958307fc83cc56f486453885e9426cd88f93d80/controllers/tf_controller.go#L439)) using the reconciledTerraform variable.

Resolution:
The fix involves modifying the defer statement so it no longer updates the metric. This ensures that the metric updates performed through the reconciliation code in r.RecordReadiness are preserved and not overwritten by outdated data.

By merging this PR, the metrics accurately reflect the current state of Terraforms, thereby improving our system's reliability of status reporting.

### DIFF
--- a/controllers/tf_controller.go
+++ b/controllers/tf_controller.go
@@ -136,7 +136,6 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 
 		// Record Prometheus metrics.
-		r.Metrics.RecordReadiness(ctx, &terraform)
 		r.Metrics.RecordSuspend(ctx, &terraform, terraform.Spec.Suspend)
 		r.Metrics.RecordDuration(ctx, &terraform, reconcileStart)
 	}()


controllers/tf_controller.go
```diff
@@ -136,7 +136,6 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		}
 
 		// Record Prometheus metrics.
-		r.Metrics.RecordReadiness(ctx, &terraform)
 		r.Metrics.RecordSuspend(ctx, &terraform, terraform.Spec.Suspend)
 		r.Metrics.RecordDuration(ctx, &terraform, reconcileStart)
```
yitsushi, line 140: Are these safe from this? They care using the same reference, or are those functions check extra criteria that wasn't checked in `RecordReadiness` so they are not affected by the same issue?

TarasLykhenko: Yes, these functions check extra criteria that aren't checked in the RecordReadiness, so they are not affected by the overwrite issue.

TarasLykhenko:
@yitsushi, the e2e test is failing, but it seems unrelated to this change. What needs to be done to unblock it?

ilithanos:
@TarasLykhenko #1290 should fix the e2e tests when merged.

ricardomiguel-os:
Hi @yitsushi , is it possible to merge this change? thank you

yitsushi:
Ah the rebase/merge invalidated the approval 🤦‍♀️

yitsushi (approved):

chanwit (approved):
LGTM :1st_place_medal: 

Thank you @TarasLykhenko for the fix!

ricardomiguel-os:
@yitsushi, this action failed to push the image.

https://github.com/flux-iac/tofu-controller/actions/runs/8949941026/job/24584776720

Do you think that a retry might solve the issue? Thanks

yitsushi:
> @yitsushi, this action failed to push the image.
> 
> https://github.com/flux-iac/tofu-controller/actions/runs/8949941026/job/24584776720
> 
> Do you think that a retry might solve the issue? Thanks

hmm, it seems it should, unless they removed `alpine:3.19` which doesn't seem reasonable.

> ERROR: failed to solve: alpine:3.19: failed to resolve source metadata for docker.io/library/alpine:3.19: failed to do request: Head "https://registry-1.docker.io/v2/library/alpine/manifests/3.19": EOF

ricardomiguel-os:
@yitsushi, do you know when the next RC version will be released?

I'm just asking because version v.0.14.1 had some critical vulnerabilities, and the current one has this metrics bug. Thanks

yitsushi:
No, I don't know.

PR #1302, opened 2024-05-31
Speedup wait for pod ip

akselleirv:
The initial startup of a pod took atleast 15 seconds when waiting for an pod IP assignment. This change uses a watch on the pod instead of a polling mechanism which results in a much faster startup time. 

The only downside is that the clientset must be defined in the TerraformReconciler, however, the dependency is already existing. Improvement is noticeable in the e2e test where it perviously took 15 min where it now took 10 min.


ilithanos (approved):
Nice find :)

PR #1310, opened 2024-06-04
Upgrade go version to 1.22 and set version via go.mod in workflow

akselleirv:
Version upgrade of controller-gen was needed in order to fix issue after the Go version upgrade. The controller-gen upgrade resulted in reformatting of the CRD.


ilithanos:
It looks good to me, and will also solve one of the issues I had earlier with the api stuff running an older go version than the rest.

I still don't know why that was pinned to go 1.19, but it seemed like someone had a reason to do it based on the documentation.

@chanwit @yitsushi was there a reason or could we upgrade.

Looks like all tests etc are parsing.

chanwit (approved):

chanwit:
Upgrading to v1.22 is fine.

Thank you both!

PR #1334, opened 2024-06-06
Bump controller-runtime and k8s.io/*

akselleirv:


go.mod
```diff
@@ -200,10 +204,6 @@ replace (
 	// v0.15.1-0.20220809152546-4850a69faedb is actually the v0.16.1a tag of the fork
 	github.com/hashicorp/terraform-exec v0.16.1 => github.com/tf-controller/terraform-exec v0.15.1-0.20220809152546-4850a69faedb
 
-	k8s.io/api => k8s.io/api v0.28.4
-	k8s.io/apimachinery => k8s.io/apimachinery v0.28.4
-	k8s.io/client-go => k8s.io/client-go v0.28.4
-
```
akselleirv, line 206: Remove the pinned version. Please let me know if this should be reverted.

ilithanos: It's generally a bad idea not to have pinned versions. any specific reason you want to unpin the version?

akselleirv: Just to avoid having to manually bump the version during a Dependabot PR.

ilithanos: Not pinning it could create issues where a build is tested with a specific version, but an becomes available between that test and the release, creating a release that we can't be sure works as expected. 

I'd rather have it bumped on dependabot PR's for that reason than risk a release with an untested dependency version.

akselleirv: I might be confusing you with my initial comment @ilithanos, but it is still pinned to a version in the `require` section of the go.mod file. This only removes the overwrite of the version in the `replace`section.

ilithanos (approved):

PR #1379, opened 2024-07-08
Add .tfstate.disablePlanLock

### LINKED ISSUE #1363: Support lock=false for terraforms in drift-detection-only mode
Hello!

Seems like there is a handy feature missing in the tofu controller.

In our case for some resources we use it in drift-detection-only mode and from time to time, there is a conflict between the external tool and tofu-controller.

So we are interested in setting  lock=false for resources in drift-detection-only mode

hirenko-v:
Fixes https://github.com/flux-iac/tofu-controller/issues/1363
This PR introduces `.tfstate.disablePlanLock` which will make possible to disable state lock for the plan stage.
It is useful when `tofu-controller` in drift-detection or plan-only mode is used in pair with another tool. The new feature helps to avoid lock conflicts


akselleirv:
Hello @hirenko-v, can you please rebase on main to fix the failing tests.

hirenko-v:
@akselleirv done

alexandermarston:
Closing in favour of #1812

PR #1438, opened 2024-09-04
chore: update alpine & tf versions

mloiseleur:
# What does it do ?

1. Update alpine to 3.20
2. Introduce build for latest 1.6.x, 1.7.x, 1.8.x and 1.9.x

# Motivation

I'm aware user can build their own runners, as documented [here](https://flux-iac.github.io/tofu-controller/use-tf-controller/build-and-use-a-custom-runner-image/).
I'm quite sure I'm not the only wanting to get a more recent version by default. Until there is a switch towards tofu-runner, I suggest with this PR to provide them.

# Notes

I kept the logic I have seen in https://github.com/flux-iac/tofu-controller/pull/1218 or https://github.com/flux-iac/tofu-controller/pull/528
The [upstream cli policy](https://support.hashicorp.com/hc/en-us/articles/360021185113-Support-Period-and-End-of-Life-EOL-Policy) is not that wide. If maintainer prefers, the scope of this PR can be reduced to only provides the three latest TF version (1.9, 1.8 and 1.7).

### DIFF
--- a/.github/workflows/release-runners.yaml
+++ b/.github/workflows/release-runners.yaml
@@ -75,7 +75,7 @@ jobs:
       packages: write # needed for ghcr access
     strategy:
       matrix:
-        tf_version: [1.0.11, 1.1.9, 1.2.9, 1.3.9, 1.4.6, 1.5.5]
+        tf_version: [1.0.11, 1.1.9, 1.2.9, 1.3.9, 1.4.6, 1.5.7, 1.6.6, 1.7.5, 1.8.5, 1.9.5]
     steps:
       - name: Check out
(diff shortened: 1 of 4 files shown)


ilithanos:
We can't default to the newest versions of terraform due to BSL Licensing in terraform, which is why we are still on the old terraform version. 

So if you want to run a BSL licensed version of terraform with the TF controller you will have to run on a custom image as it is right now at least. 

So for now there are some dockerfiles available in the following repo:

https://github.com/flux-iac/tf-runner-images

To help you build specific container images for specific terraform versions.

PR #1459, opened 2024-10-12
README.md: remove roadmap

akselleirv:
While having a roadmap is great, circumstances have changed, and we can no longer commit to a specific timeline. It might be better to remove the roadmap for now.


ilithanos (approved):

ilithanos:
I completely agree with this, since it’s now completely community maintained committing to a roadmap like this just isn’t realistic

PR #1460, opened 2024-10-13
Added exponential backoff on reconciliation failure

### LINKED ISSUE #1335: Exponentially backoff on reconciliation failure
Currently it is only possible to define a retry with `spec.retryInterval` and set the value to a static of amount of time.

However, in case of multiple Terraform resources which fails at the same time this can create a lot of noise and will constantly retry based on the provided retry interval. A more graceful approach could be to add exponentially backoff of the retry interval.

A new field named `spec.retryStrategy` can be introduced and the default value would be `StaticInterval` to keep it backward compatible, or the user can choose `ExponentialBackoff`. The first retry would be after 15 seconds and the next one at 30 seconds etc. and then set a maximum requeue time.

akselleirv:
Closes #1335


ilithanos (approved):
Approved

PR #1502, opened 2025-01-13
Upgrade Go version to 1.23.4 and deps

akselleirv:


yitsushi (approved):
lgtm, I don't see any breaking parts, but didn't test manually.

PR #1507, opened 2025-01-15
fix: goreleaser v1 to v2

akselleirv:


chanwit (approved):
Thank you for taking care of this @akselleirv

PR #1519, opened 2025-02-06
fix(oci): allow unlimited layer size

### LINKED ISSUE #1511: Error when OCI Layers uncompressed size is bigger than 100MB
Hi all,

thanks for the amazing work.

I want to run the controller in a network restricted area an cannot download providers and so on on the fly. This is why I wanted to put everything into OCI and give that to the runner. Due to the size of the terraform kubernetes provider a OCI layers compressed size is 44MB but its uncompressed size is 134MB.

This results in the following error on the Terraform CRD:

```txt

rpc error: code = Unknown desc = failed to untar artifact, error: tar ".terraform/providers/registry.opentofu.org/hashicorp/kubernetes/2.35.1/linux_amd64/terraform-provider-kubernetes" is bigger than max archive size of 104857600 bytes   13m

```

The default size of the buffer is set [here](https://github.com/fluxcd/pkg/blob/main/tar/tar.go#L29).
There error is thrown [here](https://github.com/flux-iac/tofu-controller/blob/f4893afda8d7e1dec36299216852bc9fef39e3ad/runner/server.go#L81).
I didnt find a standard way in this controllers code to increase the limit to, lets say 4GB (given the pods mem is that high).

Is that something which could already be configured and I have overlooked it or is it something I could work on an PR it?

My runner version is ghcr.io/flux-iac/tf-runner:v0.16.0-rc.5

Thanks in advance.

KarstenSiemer:
This fix gathers the value for unlimited from the tar lib and passes it to the untar function to allow layer sizes bigger than the default of 100MB.

fixes #1511 as discussed.
Currently I did not implement tests. If this is wanted, I'll gladly add them if you would be okay with v1beta2 of `github.com/fluxcd/source-controller/api`.

I have this code running now with OCIs as big as 500MB for about 2 weeks. It is running flawless, I am really happy with this controller. Thanks for all the efforts, this makes a huge difference for me :-)

Thanks for your consideration.

### DIFF
--- a/controllers/tc000012_src_ocirepository_no_outputs_test.go
+++ b/controllers/tc000012_src_ocirepository_no_outputs_test.go
@@ -134,7 +134,6 @@ func Test_000012_src_ocirepository_no_outputs_test(t *testing.T) {
 			return nil
 		}
 		for _, c := range createdHelloWorldTF.Status.Conditions {
-			fmt.Printf("Conditions: %+v \n", c)
 			if c.Type == "Plan" {
 				return map[string]interface{}{
 					"Type":    c.Type,
--- a/controllers/tc000012_src_ocirepository_no_outputs_test.go
+++ b/controllers/tc000012_src_ocirepository_no_outputs_test.go
@@ -1,0 +1,235 @@
+package controllers
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"testing"
+	"time"
+
+	infrav1 "github.com/flux-iac/tofu-controller/api/v1alpha2"
+	. "github.com/onsi/gomega"
+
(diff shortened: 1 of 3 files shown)


akselleirv:
Hello @KarstenSiemer, thank you for your contribution and the research you have done to verify that this works as intended. Yes, please feel free to add tests with `OCIRepository` using the v1beta2 version.

akselleirv (approved):
LGTM! Thank you very much @KarstenSiemer 🥇

Glad to hear that you enjoyed working on this contribution! I hope you consider making further contributions to this project.

(I'll fix the failing trivy job on main)

PR #1527, opened 2025-02-13
chore: upgrade to go 1.24 + upgrade deps

akselleirv:


internal/informer/branch-planner/informer.go
```diff
@@ -148,6 +148,12 @@ func (i *Informer) updateHandler(oldObj, newObj interface{}) {
 
 	ctx := context.Background()
 
+	// The informer cache does not update when the status field is updated, so we need to get the object from api server.
+	if err := i.client.Get(ctx, client.ObjectKeyFromObject(new), new); err != nil {
+		i.log.Info("failed getting object", "namespace", new.Namespace, "name", new.Name, "error", err)
+		return
+	}
+
```
akselleirv, line 156: A race condition occured when updating the deps. Was a problem with the update handler in the branch-planner working on a stale object from the informer cache. The status field of the object was is not updated by the informer, from my understanding, that resulted in the failed test.

Therefore I changed it to fetching the object from the API server instead of using the cached object. Consequence of the change is the extra API call.

PR #1549, opened 2025-03-19
Rename Weave GitOps to tofu-controller

pyaillet:
Hi and thanks for the project!

I found the references to the old repo a bit disturbing while browsing the docs.

So here is a small PR to change references to the actual repo.


akselleirv (approved):
Thank you, LGTM!

PR #1602, opened 2025-09-08
chore: upgrade controller-runtime to v0.22.0 and their dependencies

ricardo-mng:
The goal of this PR is to resolve the issue of Terraform CRs drifting reconciliations blocking the reconciliation of new CRs when a large number of Terraform CRs are updated within a specific period (configuration of `spec.interval`).

On controller-runtime [v0.22.0](https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.22.0), the default behaviour is to give low priority to CRs reconciliation when they are unchanged on all add functions `WithLowPriorityWhenUnchanged`

To upgrade controller-runtime, I had to upgrade some packages, including a new [version](https://github.com/fluxcd/pkg/pull/1022) of fluxcd/pkg/ssa.

To include this new version of this project, I was advised (by a Flux maintainer) to also upgrade the other FluxCD package versions, as per this go.mod file from source-controller: https://github.com/fluxcd/source-controller/blob/v1.7.0-rc.3/go.mod


alexandermarston:
@ricardo-mng would you mind rebasing against main?

PR #1661, opened 2025-12-05
fix: clear terraform state lock condition

alexandermarston:
I noticed that we weren't clearing the LockHeld condition, even after a subsequent successful plan, apply, etc which was leaving us with some weird looking conditions.

```yaml
  conditions:
  - lastTransitionTime: "2025-12-04T17:22:15Z"
    message: 'No drift: v0.6.0@sha1:db390e773761318c97c6363f1432bf3bfff5efb2'
    observedGeneration: 4
    reason: NoDrift
    status: "True"
    type: Ready
  - lastTransitionTime: "2025-11-05T17:31:17Z"
    message: Outputs available
    reason: TerraformOutputsAvailable
    status: "True"
    type: Output
  - lastTransitionTime: "2025-11-05T17:31:16Z"
    message: Plan no changes
    reason: TerraformPlannedNoChanges
    status: "False"
    type: Plan
  - lastTransitionTime: "2025-11-07T15:06:58Z"
    message: 'Terraform Locked with Lock Identifier: 2b8d8dc3-417c-8df9-e692-9b5a69d643cd'
    reason: LockHeld
    status: "True"
    type: StateLocked
```

### DIFF
--- a/api/v1alpha2/terraform_types.go
+++ b/api/v1alpha2/terraform_types.go
@@ -628,6 +628,8 @@ func TerraformApplied(terraform *Terraform, revision string, message string, isD
 		terraform.Status.Inventory = &ResourceInventory{Entries: entries}
 	}
 
+	TerraformStateLockReleased(terraform)
+
 	SetTerraformReadiness(terraform, metav1.ConditionUnknown, TFExecApplySucceedReason, message+": "+revision, revision)
 	return terraform
 }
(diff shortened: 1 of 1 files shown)


api/v1alpha2/terraform_types.go
```diff
@@ -766,6 +776,15 @@ func TerraformStateLocked(terraform *Terraform, lockID, message string) *Terrafo
 	return terraform
 }
 
+// TerraformStateLockReleased clears any state lock condition and resets lock tracking fields.
+func TerraformStateLockReleased(terraform *Terraform) *Terraform {
+	conditions.Delete(terraform, ConditionTypeStateLocked)
+	terraform.Status.Lock.Pending = ""
+	terraform.Status.Lock.LastApplied = ""
+
+	return terraform
```
artem-nefedov: nit: Return value is never used, so maybe function should return anything? Or better yet, make it a method, since that IMO makes more sense when you only need to mutate the object.

PR #1668, opened 2025-12-08
fix: reconcile terraform when source revision changes

### LINKED ISSUE #1667: Tofu controller not reconciling a Terraform CR when GitRepository has a new revision
Hello! First, so excited about the accelerated pace and work going into this tool! Thank you.

I recently updated from 0.16.5 to 0.16.7, and I think that some of the changes caused an issue where a Terraform with a GitRepository source  won't reconcile even if a new GitRepository revision is present.

To reproduce, add a simple Terraform CR with GitRepository source, update the git repository, notice that the GitRepository points at a new revision, and the Terraform resource staying on the previous one.

If need more info I'll gladly share whatever needed, and if you'd like I can try to contribute a PR, but since it may be related to recent changes of the smarter-reconciliation around #1648 and #1662 you'll probably already have a fix in mind.

alexandermarston:
Closes #1667

### DIFF
--- a/controllers/tf_controller.go
+++ b/controllers/tf_controller.go
@@ -163,19 +163,10 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	// Return early if the Terraform is suspended.
-	traceLog.Info("Check if the Terraform resource is suspened")
+	traceLog.Info("Check if the Terraform resource is suspended")
 	if terraform.Spec.Suspend {
 		log.Info("Reconciliation is suspended for this object")
 		return ctrl.Result{}, nil
-	}
-
-	// Check whether we need to reconcile the release at this time
-	if shouldReconcile, requeueAfter := r.shouldReconcile(terraform); !shouldReconcile {
-		log.Info("Skipping reconciliation. Interval has not elapsed since last plan",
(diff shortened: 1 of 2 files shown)


ricardo-mng (approved):

controllers/tf_controller.go
```diff
@@ -256,6 +247,22 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		conditions.MarkUnknown(terraform, meta.ReadyCondition, meta.ProgressingReason, "Reconciliation in progress")
 	}
 
+	// Check whether we need to reconcile the release at this time
+	if shouldReconcile, reason, requeueAfter := r.shouldReconcile(terraform, sourceObj); !shouldReconcile {
+		log.Info("Skipping reconciliation. Interval has not elapsed since last plan",
+			"lastPlanAt", terraform.Status.LastPlanAt,
+			"nextReconcile", terraform.Status.LastPlanAt.Add(terraform.Spec.Interval.Duration),
+			"requeueAfter", requeueAfter,
+			"reason", reason)
```
artem-nefedov: @alexandermarston  The function only sets non-empty reason when it returns `true`, but this log is only printed when function returned `false`, so it seems that the reason will never actually be used?

PR #1670, opened 2025-12-09
fix: delete reconciling condition on successful initial apply

### LINKED ISSUE #1669: Reconciling condition not cleared on first apply
It is the same problem as #1657.  The fix from #1658 did help in most cases, but not in all of them.
Specifically, the problem with Reconciling and Ready condition being present together still happens when both of the following conditions are true:

1. It's a new apply (there's no saved state yet)
2. Terraform doesn't write outputs

Example of resulting status (tested on 1000a3f768d211be850dcd365d1b550eda1f862b, currently latest):

```yaml
status:
  conditions:
  - lastTransitionTime: "2025-12-09T17:06:14Z"
    observedGeneration: 1
    reason: Progressing
    status: "True"
    type: Reconciling     # This shouldn't be here!
  - lastTransitionTime: "2025-12-09T17:22:11Z"
    message: 'Applied successfully: <redacted>'
    observedGeneration: 1
    reason: TerraformAppliedSucceed
    status: "True"
    type: Ready
  - lastTransitionTime: "2025-12-09T17:22:06Z"
    message: Applied successfully
    observedGeneration: 1
    reason: TerraformAppliedSucceed
    status: "True"
    type: Apply
  - lastTransitionTime: "2025-12-09T17:06:56Z"
    message: Plan generated
    observedGeneration: 1
    reason: TerraformPlannedWithChanges
    status: "True"
    type: Plan
```

Second and all subsequent applies will clear the Reconciling condition, but it being present after first one greatly slows down pipelines relying on Kustomization healthchecks, especially if Terraform reconcile interval is long.

artem-nefedov:
fixes #1669

The problem with initial apply without outputs is that it only calls SetTerraformReadiness from [here](https://github.com/flux-iac/tofu-controller/blob/v0.16.0-rc.7/api/v1alpha2/terraform_types.go#L631), and because it's called with ConditionUnknown, Reconciling condition is not deleted there. After that, it sets Ready condition to true in a different place in a convoluted matter (this part probably needs to be rewritten eventually).


controllers/tf_controller_reconcile.go
```diff
@@ -309,6 +310,7 @@ func (r *TerraformReconciler) reconcile(ctx context.Context, patchHelper *patch.
 		if cond.Reason == infrav1.PlannedWithChangesReason && strings.HasPrefix(cond.Message, "Plan generated") {
 			// do nothing
 		} else if cond.Reason != meta.ProgressingReason {
+			conditions.Delete(terraform, meta.ReconcilingCondition)
```
artem-nefedov: This will shift indices, need to use a different approach

alexandermarston:
Thank you for the PR!

I don't think we have a test case that covers a first apply with no outputs, do you think you would be willing to create one? You could do something like [this](https://github.com/flux-iac/tofu-controller/blob/main/controllers/tc000010_no_outputs_test.go#L193) to assert that we have the expected conditions at the end.

alexandermarston (approved):
Thank you for the test! LGTM 🚀

Issue #1673, opened 2025-12-12
Simplify controller image


artem-nefedov:
Controller image has no reason to be built on top of [alpine](https://github.com/flux-iac/tofu-controller/blob/v0.16.0-rc.7/Dockerfile#L38), nor does it need [all these extra packages](https://github.com/flux-iac/tofu-controller/blob/v0.16.0-rc.7/Dockerfile#L44). It all comes as a baggage all the way from 415c8b4d10193882f9a1453b6bc6e55cc36bd2cb, when the controller image was built with Terraform binary in it. Terraform was already removed from the image since then, but alpine OS and all the other packages are still present to this day for no practical reason other than maybe manual debugging (which is better achieved via `kubectl debug`). This has mutliple downsides, such as larger image size and higher security risks.

Go produces static binaries that can often be easily put into `FROM scratch` image. If that is not an option, one should use a distroless image ([option 1](https://github.com/GoogleContainerTools/distroless), [option 2](https://images.chainguard.dev/directory/image/static/versions) - used by [ko](https://ko.build/) as default image). There seem to be no good reason to use [tini](https://github.com/flux-iac/tofu-controller/blob/v0.16.0-rc.7/Dockerfile#L59) as init process for the controller, since it doesn't spawn any children.

Note: This is only related to controller image, not the runner.

artem-nefedov:
Yes, exactly.

PR #1674, opened 2025-12-12
feat: use a distroless image for tofu-controller

alexandermarston:
Implements suggestions from #1673

### DIFF
--- a/planner.Dockerfile
+++ b/planner.Dockerfile
@@ -3,7 +3,11 @@ FROM --platform=$BUILDPLATFORM golang:${GO_VERSION} as builder
 
 WORKDIR /build
 
-COPY . .
+COPY go.mod go.sum ./
+COPY api/ api/
+COPY cmd/branch-planner cmd/branch-planner
+COPY internal/ internal/
+COPY tfctl/ tfctl/
 
 ARG BUILD_SHA
 ARG BUILD_VERSION


Dockerfile
```diff
@@ -1,59 +1,42 @@
...
-FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
+ARG GO_VERSION=1.25
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION} as builder
 
-ARG BUILD_SHA
-ARG BUILD_VERSION
+WORKDIR /build
 
-RUN apt-get update && apt-get install -y unzip
+COPY . .
```
TarasLykhenko: `COPY . .` works, but it can make caching less efficient since changes to unrelated files will trigger a rebuild. I would prefer to keep the original behavior.

config/manager/manager.yaml
```diff
@@ -26,9 +26,7 @@ spec:
         image: flux-iac/tofu-controller
         imagePullPolicy: IfNotPresent
         command:
-        - /sbin/tini
-        - --
-        - tofu-controller
+        - /tofu-controller
```
TarasLykhenko: The deployment currently sets command: `/tofu-controller`. Since the container’s ENTRYPOINT is already set to the same binary, do we actually need it?

Dockerfile
alexandermarston: Reverted to the old behaviour.

planner.Dockerfile
```diff
@@ -1,57 +1,42 @@
-# Build the manager binary
-FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
+ARG GO_VERSION=1.25
+FROM --platform=$BUILDPLATFORM golang:${GO_VERSION} as builder
+
+WORKDIR /build
+
+COPY . .
```
artem-nefedov: COPY change should be reverted for branch planner too?

PR #1675, opened 2025-12-12
Migrate Tofu Controller to OpenTofu

alexandermarston:
For some time, Tofu Controller has only supported up to version `1.5.7` of Terraform due to the licensing changes Hashicorp announced and implemented in 2023. As we have not been able to update to a newer version of Terraform and had not completed the migration to OpenTofu, users have been forced to deploy their own custom runner images.

This PR starts the process of moving away from Hashicorp Terraform and to the [OpenTofu](https://opentofu.org/) implementation.

For this change, a few changes have been made:

- GitHub workflows now setup OpenTofu binaries, rather than Terraform.
- Various Test Cases have been updates to account for OpenTofu specific CLI output.
- Runner Images have been updated to make the `tofu` binary available to the runner, pulling this in from the OpenTofu minimal images rather than downloading architecture specific binaries.
- Tofu Controller (and the runner) will now try to use the `terraform` binary by default, but will fall back to `tofu` if it cannot be found. As we only ship the `tofu` binary in the runner base image, this should be the default for standard deployments of Tofu Controller.

### DIFF
--- a/docs/use-tf-controller/build-and-use-a-custom-runner-image.md
+++ b/docs/use-tf-controller/build-and-use-a-custom-runner-image.md
@@ -16,7 +16,7 @@ You need Docker and Git to build the image.
 
     ```Dockerfile
     ARG BASE_IMAGE
-    ARG TOFU_VERSION=1.6.2
+    ARG TOFU_VERSION=1.11.3
 
     FROM ghcr.io/opentofu/opentofu:${TOFU_VERSION}-minimal AS tofu
 
@@ -34,7 +34,7 @@ You need Docker and Git to build the image.
 
     ```bash
     export TF_CONTROLLER_VERSION=v0.16.0-rc.7
-    export TOFU_VERSION=1.6.2
+    export TOFU_VERSION=1.11.3
     export BASE_IMAGE=ghcr.io/flux-iac/tf-runner:${TF_CONTROLLER_VERSION}-base
     export REMOTE_REPO=ghcr.io/my-org/custom-runnner
     docker build \
--- a/.github/workflows/build-and-publish.yaml
+++ b/.github/workflows/build-and-publish.yaml
@@ -27,7 +27,7 @@ jobs:
           go-version-file: go.mod
       - name: Setup OpenTofu
         run: |
-          export TOFU_VERSION=1.6.2
+          export TOFU_VERSION=1.11.3
           wget https://github.com/opentofu/opentofu/releases/download/v${TOFU_VERSION}/tofu_${TOFU_VERSION}_linux_amd64.zip
           unzip -q tofu_${TOFU_VERSION}_linux_amd64.zip tofu
           mv tofu /usr/local/bin
--- a/.github/workflows/targeted-test.yaml
+++ b/.github/workflows/targeted-test.yaml
@@ -23,7 +23,7 @@ jobs:
             **/go.mod
       - name: Setup OpenTofu
         run: |
-          export TOFU_VERSION=1.6.2
+          export TOFU_VERSION=1.11.3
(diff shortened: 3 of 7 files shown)


artem-nefedov:
This probably needs to be highlighted as a breaking change for those who build their own runner images and update terraform binary there

controllers/tf_controller_backend.go
```diff
@@ -181,10 +181,10 @@ terraform {
 
 	lookPathReply, err := runnerClient.LookPath(ctx,
 		&runner.LookPathRequest{
-			File: "terraform",
+			File: "tofu",
```
OS-alexandrebrito: Maybe add a fallback to terraform or parametrize so it doesn't become a breaking change.

alexandermarston: Thanks. I think some of your concerns will have been addressed with the latest update - we'll now fallback to the `terraform` binary if we can't find `tofu`.

Makefile
```diff
@@ -225,14 +225,25 @@ kustomize: ## Download kustomize locally if necessary.
...
+PROTOC_V ?= 33.4
+PROTOC_URL := https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_V}/protoc-${PROTOC_V}
 .PHONY: protoc
 protoc: ## Download protoc locally if necessary.
 	# download and unzip protoc
 	mkdir -p $(PROJECT_DIR)
-	curl -qLO https://github.com/protocolbuffers/protobuf/releases/download/v$(PROTOC_V)/protoc-$(PROTOC_V)-linux-x86_64.zip
-	unzip -q -o protoc-$(PROTOC_V)-linux-x86_64.zip bin/protoc -d $(PROJECT_DIR)
-	rm protoc-$(PROTOC_V)-linux-x86_64.zip
+	if [ "$(shell uname)" == "Darwin" ]; then \
```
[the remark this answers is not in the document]
artem-nefedov: It works due to transparent Rosetta 2 emulation for x86_64 binaries, but is somewhat less performant than running native arm64 ones.
In the end, any option is fine, it's not super important.

controllers/tf_controller_backend.go
```diff
@@ -179,19 +179,24 @@ terraform {
...
 	lookPathReply, err := runnerClient.LookPath(ctx,
 		&runner.LookPathRequest{
-			File: "terraform",
-		})
+			Files: []string{"terraform", "tofu"},
+		},
+	)
 	if err != nil {
-		err = fmt.Errorf("cannot find Terraform binary: %s in %s", err, os.Getenv("PATH"))
+		err = fmt.Errorf("cannot find Terraform or OpenTofu binary: %s in %s", err, os.Getenv("PATH"))
```
ricardomiguel-os: Can you change the message to something like "cannot find any of the required binaries (terraform, tofu) in PATH: %s" because if I'm not mistaken, the error will only reference the last binary lookup (tofu) and could be misleading.

docs/use-tf-controller/build-and-use-a-custom-runner-image.md
```diff
@@ -34,14 +33,13 @@ You need Docker and Git to build the image.
 2. Build the image from the directory containing the `Dockerfile` you created above:
 
     ```bash
-    export TF_CONTROLLER_VERSION=v0.16.0-rc.3
-    export TF_VERSION=1.5.7
+    export TF_CONTROLLER_VERSION=v0.16.0-rc.7
+    export TOFU_VERSION=1.6.0
```
ricardomiguel-os: We are migrating to TOFU_VERSION=1.11.1

ricardo-mng (commented):
LGTM my only concern is the tofu version, we should probably first migrate to version 1.6.2

alexandermarston:
We will have to upgrade to a version newer than 1.6.2 because of the vulnerabilities. Deciding what version.

PR #1730, opened 2026-03-14
fix: branch planner plans against main instead of PR branch when using Flux Operator

maruina:
## What
Make Branch Planner PR sources resolve only the pull request's head branch.

## Why
When Flux Operator translates `FluxInstance.sync.ref: refs/heads/main`, it creates a `GitRepository` with `spec.ref.name`. Flux Source Controller gives `name` precedence over `branch`, so Branch Planner's previous deep copy caused PR plans to run against `main`.

The generated PR-specific `GitRepository` now replaces `spec.ref` with a branch-only reference. This preserves the rest of the source configuration while ensuring the PR branch is selected.

## Tests
- `go test ./internal/server/polling`
- Added regression coverage for sources using `spec.ref.name`.

### DIFF
--- a/internal/server/polling/terraform.go
+++ b/internal/server/polling/terraform.go
@@ -161,11 +161,8 @@ func (s *Server) reconcileSource(ctx context.Context, tfName string, originalSou
 
 		spec := originalSource.Spec.DeepCopy()
 
-		if spec.Reference != nil {
-			spec.Reference.Branch = branch
-		} else {
-			spec.Reference = &sourcev1.GitRepositoryRef{Branch: branch}
-		}
+		// A PR-specific GitRepository must resolve only the PR branch.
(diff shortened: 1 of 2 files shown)


ishioni:
Can a maintainer please incorporate this? Flux-operator is all the rage now, and this bug locks out anyone using it

internal/server/polling/terraform.go
alexandermarston: I think we would be able to remove the if/else block entirely and just have the below, what do you think?

```go
spec.Reference = &sourcev1.GitRepositoryRef{Branch: branch}
```

maruina: @alexandermarston good point, I applied your suggestion

maruina:
@alexandermarston is there anything else you want me to address to be able to merge this?

PR #1744, opened 2026-04-02
feat(tfctl): add --all flag to 'reconcile' command for bulk reconciliation

cschindlbeck:
Adds --all to tfctl reconcile for bulk resource reconciliation


cmd/tfctl/main.go
```diff
@@ -127,20 +129,47 @@ func buildUninstallCmd(app *tfctl.CLI) *cobra.Command {
...
+				if len(args) > 0 {
+					return errors.New("cannot use --all and provide resource name at the same time")
+				}
+			} else {
+				if len(args) == 0 {
+					return errors.New("resource name required unless --all is specified")
+				}
+				resource = args[0]
+				if len(args) > 1 {
+					return errors.New("only one resource name accepted")
```
artem-nefedov: If command now can work on multiple resources with `--all`, it would probably make sense to also support multiple resource names

alexandermarston:
Thank you for the contribution! It looks like we've got a few vulnerabilities which are blocking the pipeline - I'll get these cleared up and then review this.

cschindlbeck:
@alexandermarston thanks a bunch, can you decide if we should include commit f2dfa70417270d7e9db7bbb98f696a353eeeddae or remove, that means, should the -A flag be included, this would be consistent with the other commands, i would revert it then

alexandermarston:
If we are just trying to trigger a bulk reconciliation, could we consider doing something like this? (fyi, this is what we use internally at my org)

```
for i in $(kubectl get terraforms.infra.contrib.fluxcd.io -n my-namespace | awk '{print $1}'); do;
kubectl annotate --namespace my-namespace --field-manager=flux-client-side-apply --overwrite terraform $i reconcile.fluxcd.io/requestedAt="$(date +%s)"
done
```

I prefer this way as I don't have to wait for the `Replan requested...` for each resource.

What are your thoughts? Perhaps we can add this to the docs.

artem-nefedov:
But it we already have `tfctl reconcile`, it should be usable for this purpose. It also adds `reconcile.fluxcd.io/requestedAt` annotation.
And you can call it in the loop too, but I think the point of this specific PR is to make `reconcile` flags similar to what `suspend`/`resume` have.

cschindlbeck:
@alexandermarston 
We could do that but it pushes that logic to the user side. Adding --all keeps the workflow inside tfctl while avoiding requiring custom scripting for a reasonable operation.

From my perspective, the goal of this PR was to make tfctl reconcile consistent with other commands like suspend/resume, which already support bulk operations via flags.

We can add your suggestion to the docs too in case of someone wants to skip the Reconcile requested. Or add another flag --skip-reconcile

PR #1754, opened 2026-04-10
fix: handle re-reconciling if source revision has changed while waiti…

### LINKED ISSUE #1755: Manual plan approval blocks re-plan on new Git revision
When using tofu-controller in manual plan approval mode, pushing new commits to the referenced `GitRepository` does not trigger a new plan if a previous plan is still pending approval. The Terraform resource continues to request approval for the old planId, even though the GitRepository revision is no longer HEAD.

**Environment**  
- tofu-controller: 1.6.2
- Flux: `GitRepository` + Terraform CR in `flux-system` namespace 

**Steps to reproduce**  
1. Enable manual plan approval (no `spec.approvePlan` key present).  
2. Reconcile a `Terraform` resource until it generates a plan and enters 'waiting for manual approve'.  
3. Push new commits to Git (either Terraform code changes or Terraform resource var changes) so the `GitRepository` artifact revision updates.  
4. Observe the Terraform resource status and controller behavior.

**Expected behavior**  
- A new Git revision should result in a new plan being generated, or the pending plan should be invalidated, so the user can approve a plan for the latest revision.  
- The 'approvePlan' request should correspond to the current desired revision, not an older commit.

**Actual behavior**  
- Reconciliation stops with the log message: 'reconciliation is stopped to wait for a manual approve'.  
- The Terraform resource keeps a pending plan tied to an older commit, for example:  
  - `status.plan.pending: plan-OPSEXP-3953-tofu-controller-a4be7c990b`  
  - `lastPlannedRevision` and `lastAttemptedRevision` remain `...@sha1:a4be7c990b...`  
  - Meanwhile the `GitRepository` shows a newer stored artifact revision, for example `...@sha1:8d00d452...`  
- Result: no automatic re-plan happens while a plan is pending approval, even when the source revision changes.

alexandermarston:
Closes #1755


controllers/tf_controller.go
```diff
@@ -387,6 +387,23 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
 		// case 3:
+		// if revision is changed, there's a pending plan awaiting manual approval,
+		// and the user hasn't approved or replanned, we should clear the Pending
+		// Plan to trigger re-plan with the new source revision.
+		//
+		if sourceObj.GetArtifact().Revision != terraform.Status.LastAttemptedRevision &&
+			terraform.Status.Plan.Pending != "" &&
+			!r.forceOrAutoApply(terraform) &&
+			!r.shouldApply(terraform) {
```
TarasLykhenko, line 396: From what I can see, the forceOrAutoApply check is already handled within shouldApply, so it seems redundant here and could be removed.
```suggestion
			terraform.Status.Plan.Pending != "" &&
			!r.shouldApply(terraform) {
```

alexandermarston: Good spot - thank you!

ricardo-mng (approved):

PR #1769, opened 2026-04-16
feat: add ResourceQuota support with jitter-based requeue

TarasLykhenko:
Handle Kubernetes Pod quota exhaustion errors gracefully by detecting them during runner creation and retrying with configurable jitter-based delays, preventing thundering herd scenarios when quota is tight.

## Key Features

- Detects Pod quota exhaustion errors (HTTP 403 with "exceeded quota" message)
- Automatically retries with configurable base delay + proportional jitter (default: 5s with 0.4 factor → range [5s, 7s])
- Spreads retries naturally across time to prevent thundering herd
- Feature-flagged behind `--quota-retry-enabled` (opt-in, default off)
- Emits a Kubernetes `Warning` event (`RunnerQuotaExhaustedReason`) on each retry — aggregated by the API server for clean `kubectl describe` output
- No changes to existing error handling for non-quota errors

## Changes

- `utils/quota.go` — New `IsQuotaError()` function for detecting quota exhaustion errors
- `controllers/tf_controller.go` — Add quota retry configuration fields and jitter-based requeue logic in `Reconcile()`; emit warning event on retry
- `api/v1alpha2/condition_types.go` — Add `RunnerQuotaExhaustedReason` event reason constant (following Flux convention)
- `cmd/manager/main.go` — Add `--quota-retry-enabled`, `--quota-retry-delay`, and `--quota-retry-jitter-factor` flags
- `charts/tofu-controller/values.yaml` + `templates/deployment.yaml` — Expose quota retry flags in the Helm chart
- `charts/tofu-controller/README.md` — Regenerated helm-docs
- `controllers/tc000361_runner_quota_exhausted_test.go` — Integration test for quota exhaustion handling

## Configuration

Feature disabled by default (no behaviour change for existing deployments):
```
./manager
```

Enable with defaults (5s base, up to 40% jitter → [5s, 7s]):
```
./manager --quota-retry-enabled=true
```

Slow cluster (10s base, up to 50% jitter → [10s, 15s]):
```
./manager --quota-retry-enabled=true --quota-retry-delay=10s --quota-retry-jitter-factor=0.5
```

## Test Plan

- Integration test (tc000361) covering quota exhaustion scenario with Terraform reconciliation
- Verifies runner creation failure due to quota is properly detected and requeued
- Code builds successfully
- No regressions in existing tests

## Behavior Change

**Before:** Runner creation fails with quota error → hard error propagates up, reconciliation fails immediately

**After:** Runner creation fails with quota error → detected and requeued with jitter delay, Kubernetes Warning event emitted (aggregated), reconciliation retries automatically with spread-out timing. Non-quota errors are unaffected.


controllers/tf_controller.go
```diff
@@ -420,6 +426,21 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 	if err != nil {
 		log.Error(err, "unable to lookup or create runner")
+
+		// Check if this is a quota error
+		if utils.IsQuotaError(err) {
+			// Calculate jitter (0 to QuotaRetryJitter)
+			jitter := time.Duration(rand.Intn(int(r.QuotaRetryJitter.Milliseconds()))) * time.Millisecond
+			jitteredDelay := r.QuotaRetryDelay + jitter
+
+			traceLog.Info("runner quota exhausted, retrying",
```
ricardomiguel-os, line 437: Should we set this to debug as we can have these logs in loop for some time?

controllers/tf_controller.go
```diff
@@ -420,6 +426,21 @@ func (r *TerraformReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	runnerClient, closeConn, err := r.LookupOrCreateRunner(ctx, terraform, sourceObj.GetArtifact().Revision)
 	if err != nil {
 		log.Error(err, "unable to lookup or create runner")
+
+		// Check if this is a quota error
+		if utils.IsQuotaError(err) {
+			// Calculate jitter (0 to QuotaRetryJitter)
+			jitter := time.Duration(rand.Intn(int(r.QuotaRetryJitter.Milliseconds()))) * time.Millisecond
```
ricardomiguel-os: Validate that QuotaRetryJitter is greater than zero, otherwise we can have panic errors.
```
if ms := r.QuotaRetryJitter.Milliseconds(); ms > 0 {
      jitter = time.Duration(rand.Int63n(ms)) * time.Millisecond
  }
```

Or you can use: apimachinery/pkg/util/wait.Jitter(r.QuotaRetryDelay, factor) that already handle this 0 use case

ricardo-mng (dismissed):

controllers/tf_controller.go
TarasLykhenko: Good catch, implemented the wait.Jitter sugestion.

controllers/tf_controller.go
TarasLykhenko: From the code, traceLog is used for that purpose. I think there is no need to switch to debug mode.

ricardo-mng: Ok, what about emmiting an event like we are doing on DependencyNotReadyReason?
We can create a new one RunnerQuotaExhaustedReason and this way have visibility on why the reconciliation is not progressing.

TarasLykhenko: Add the event, can you check again?

ricardo-mng (approved):

PR #1779, opened 2026-04-27
fix: use Spec.BackendConfig.Labels when set

### LINKED ISSUE #1774: changing labels on the `Terraform` CR breaks state persistence
## Summary                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                       
  When a `Terraform` CR has labels that are different than the labels on Kubernetes state Secret (for example because new labels have been added to the `Terraform` CR) `Teraform` can become **permanently stuck in an empty bootstrap state**. Every reconciliation after that the tf-runner:                                                                        
                                                                                                                                                                                                                                                                                       
  1. Runs `terraform init`, cannot find the existing Secret and persists a fresh, empty state.                                                                                                                                                                                         
  2. Runs `terraform apply` from scratch.                                                                                                                                                                                                                                              
  3. Persists real state briefly after apply completes.                                                                                                                                                                                                                                  
  4. Next cycle repeats step 1, wiping the just-written state back to empty.                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                         
  Net effect: state is lost on every reconcile. Each reconcile plans from scratch.                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                         
  The CR-level field that would logically let a user avoid this — `spec.backendConfig.labels` — is declared in the CRD and in the Go struct, but **the controller ignores it**. It always serialises the full `terraform.Labels` (the CR's `metadata.labels`) into the backend HCL       
  instead.                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                         
  ## Reproduction                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                       
  1. Create a `Terraform` CR without explicit `spec.backendConfig`. Add any custom label to `metadata.labels` (e.g. `example.com/foo: bar`).                                                                                                                                             
  2. Let tofu-controller reconcile at least once; the state Secret is created with that label set plus the backend's internal labels.                                                                                                                                                  
  3. Add a second custom label to the CR's `metadata.labels`, OR change the value of an existing custom label, OR remove one of them (any change to the set).                                                                                                                            
  4. Wait for the next reconciliation.                                                                                                                                                                                                                                                   
  5. Observe: the state Secret's `resourceVersion` increments but its `tfstate` payload always reads back as `serial=1, resources=[]` regardless of what apply did. Poll it at sub-second granularity and you'll see a brief window after each apply where real state exists, followed   
  immediately by a wipe on the next init.                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                         
  A quick way to observe the "real state briefly exists, then gets wiped" pattern with 1s polling:                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                       
  ```bash                                                                                                                                                                                                                                                                                
  while true; do                                                                                                                                                                                                                                                                       
    kubectl get secret tfstate-default-<name> -o json \                                                                                                                                                                                                                                
      | jq -r '.data.tfstate' | base64 -d | gunzip \                                                                                                                                                                                                                                     
      | jq -c '{serial, res: (.resources | length)}'                                                                                                                                                                                                                                     
    sleep 1                                                                                                                                                                                                                                                                              
  done                                                                                                                                                                                                                                                                                   
  ```                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                         
  ## Root cause                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                       
  ### 1. tofu-controller unconditionally copies all CR labels into the backend HCL                                                                                                                                                                                                       
   
  `controllers/tf_controller_backend.go`:                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                       
  - Lines 82–99 (explicit `spec.backendConfig` branch) serialise `terraform.Labels` into the HCL `labels = {...}` block.                                                                                                                                                                 
  - Lines 105–122 (no-`backendConfig` branch) do the same.                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                         
  Both call:                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                         
  ```go                                                                                                                                                                                                                                                                                
  getLabelsAsHCL(terraform.Labels, 6)                                                                                                                                                                                                                                                  
  ```                                                                                                                                                                                                                                                                                    
   
  `getLabelsAsHCL` (lines 433–444) iterates `map[string]string` and emits every key/value pair verbatim — no allowlist, no exclusion, no source selection between CR metadata labels and `Spec.BackendConfig.Labels`.                                                                    
                                                                                                                                                                                                                                                                                       
  **The `Spec.BackendConfig.Labels` field defined in the CRD and the Go struct is never read.** Setting it on the CR has no effect.                                                                                                                                                      
                                                                                                                                                                                                                                                                                       
  ### 2. The terraform kubernetes backend uses the `labels` config as a LIST selector for state reads                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                       
  From the kubernetes backend in `internal/backend/remote-state/kubernetes/client.go` (paraphrased):                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                       
  ```go                                                                                                                                                                                                                                                                                  
  func (c *RemoteClient) Get() (payload *remote.Payload, err error) {                                                                                                                                                                                                                  
      secretList, err := c.getSecrets() // LIST with LabelSelector = c.getLabels()                                                                                                                                                                                                       
      if len(secretList) == 0 {                                                                                                                                                                                                                                                          
          return nil, nil // "no state" if selector matches zero Secrets                                                                                                                                                                                                                 
      }                                                                                                                                                                                                                                                                                  
      ...                                                                                                                                                                                                                                                                                
  }                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                       
  func (c *RemoteClient) getSecrets() ([]unstructured.Unstructured, error) {                                                                                                                                                                                                             
      ls := metav1.SetAsLabelSelector(c.getLabels())                                                                                                                                                                                                                                     
      return c.kubernetesSecretClient.List(ctx, metav1.ListOptions{                                                                                                                                                                                                                      
          LabelSelector: metav1.FormatLabelSelector(ls),                                                                                                                                                                                                                                 
      })                                                                                                                                                                                                                                                                                 
  }                                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                         
  func (c *RemoteClient) getLabels() map[string]string {                                                                                                                                                                                                                               
      l := map[string]string{                                                                                                                                                                                                                                                            
          tfstateKey:             "true",                                                                                                                                                                                                                                                
          tfstateSecretSuffixKey: c.nameSuffix,                                                                                                                                                                                                                                          
          tfstateWorkspaceKey:    c.workspace,                                                                                                                                                                                                                                           
          managedByKey:           "terraform",                                                                                                                                                                                                                                           
      }                                                                                                                                                                                                                                                                                  
      maps.Copy(l, c.labels) // user-provided `labels` merged in                                                                                                                                                                                                                         
      return l                                                                                                                                                                                                                                                                           
  }                                                                                                                                                                                                                                                                                    
  ```                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                         
  When `Get` returns `nil, nil`, `backend_state.go:StateMgr` treats the state as non-existent and runs the bootstrap path: `WriteState(states.NewState()); PersistState(nil)` — which overwrites the existing Secret's `data.tfstate` with an empty state.                               
                                                                                                                                                                                                                                                                                         
  ### 3. The backend's `Put` does not update existing Secret labels                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                       
  ```go                                                                                                                                                                                                                                                                                  
  // internal/backend/remote-state/kubernetes/client.go Put(...)                                                                                                                                                                                                                       
  secret, err := c.getSecret(secretName) // Get by NAME, not by labels                                                                                                                                                                                                                   
  if err != nil {                                                                                                                                                                                                                                                                        
      if !k8serrors.IsNotFound(err) {                                                                                                                                                                                                                                                    
          return err                                                                                                                                                                                                                                                                     
      }                                                                                                                                                                                                                                                                                  
      secret = &unstructured.Unstructured{ /* with c.getLabels() */ }                                                                                                                                                                                                                    
      secret, err = c.kubernetesSecretClient.Create(...) // only on first create                                                                                                                                                                                                         
  }                                                                                                                                                                                                                                                                                      
  setState(secret, data) // touches only secret.Object["data"]                                                                                                                                                                                                                           
  _, err = c.kubernetesSecretClient.Update(ctx, secret, metav1.UpdateOptions{})                                                                                                                                                                                                          
  ```                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                         
  Once the Secret exists, `Put` only touches `data.tfstate`. **Labels on the pre-existing Secret are never rewritten.**                                                                                                                                                                  
                                                                                                                                                                                                                                                                                       
  ### Putting it together                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                       
  The sequence that produces the failure:                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                       
  1. First reconciliation of a `Terraform` CR with label set L₀. Backend HCL labels = L₀ ∪ {internal tfstate labels}. Secret created with that full label set.                                                                                                                           
  2. CR labels change to L₁ (e.g. a label added or removed, or a value modified). Backend HCL labels = L₁ ∪ {internal tfstate labels}.                                                                                                                                                 
  3. Next reconciliation:                                                                                                                                                                                                                                                                
     - `terraform init` → `StateMgr` → `Get` → `getSecrets` issues a LIST with selector L₁ ∪ {internal tfstate labels}.                                                                                                                                                                  
     - The existing Secret has labels L₀ ∪ {internal tfstate labels}. If L₁ ⊄ L₀, **the selector doesn't match** and List returns empty.                                                                                                                                                 
     - `Get` returns `nil, nil` → bootstrap path runs → empty state is persisted via `Put`, which finds the existing Secret by name and overwrites `data.tfstate` with `states.NewState()`.                                                                                              
  4. Plan then shows every resource as "create" (because state is empty). Apply runs, creates real resources, persists state briefly.                                                                                                                                                    
  5. Next reconciliation repeats from step 3. Forever.                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                         
  This is permanent until either the state Secret is manually deleted (so terraform recreates it with the new full label set) or all non-internal labels are removed from the CR.                                                                                                        
                                                                                                                                                                                                                                                                                         
  ### Evidence from a live cluster                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                       
  On a cluster exhibiting this bug we observed state Secret transitions at 1-second polling granularity:                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                       
  ```text                                                                                                                                                                                                                                                                                
  t=2s   serial=1 res=0 size=181B   # post-init bootstrap                                                                                                                                                                                                                              
  t=36s  serial=2 res=1 size=1697B  # apply persisted real state                                                                                                                                                                                                                         
  t=38s  serial=1 res=0 size=181B   # next cycle's init wiped it                                                                                                                                                                                                                         
  ```                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                         
  The "empty → real → empty" cycle repeats on every reconcile interval, and each real-state window coincides with the creation of new resources by the plan's "create" actions. Because the state is empty every time planning runs, the controller never recognises that those resources
   already exist and creates fresh ones on each pass.                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                         
  Label comparison on a reproducing cluster:                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                       
  ```text                                                                                                                                                                                                                                                                                
  Secret.metadata.labels (set at Secret creation):                                                                                                                                                                                                                                     
    app.kubernetes.io/managed-by=terraform                                                                                                                                                                                                                                               
    tfstate=true                                                                                                                                                                                                                                                                         
    tfstateSecretSuffix=...                                                                                                                                                                                                                                                              
    tfstateWorkspace=default                                                                                                                                                                                                                                                             
    kustomize.toolkit.fluxcd.io/name=...                                                                                                                                                                                                                                                 
    kustomize.toolkit.fluxcd.io/namespace=...                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                         
  Terraform CR.metadata.labels (current; source for the backend HCL):                                                                                                                                                                                                                    
    example.com/some-later-added-label=...   # not on the Secret                                                                                                                                                                                                                         
    kustomize.toolkit.fluxcd.io/name=...                                                                                                                                                                                                                                                 
    kustomize.toolkit.fluxcd.io/namespace=...                                                                                                                                                                                                                                            
  ```                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                       
  The selector issued by the backend includes `example.com/some-later-added-label=...`, which the existing Secret doesn't carry → empty List → bootstrap wipe.                                                                                                                           
                                                                                                                                                                                                                                                                                       
  ## Why `spec.backendConfig.labels` does not help                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                       
  The CRD schema and the Go type `BackendConfigSpec` both declare a `labels map[string]string` field. A user looking at this naturally assumes it overrides which labels end up in the backend HCL. It does not.                                                                         
                                                                                                                                                                                                                                                                                       
  In `controllers/tf_controller_backend.go`, both the explicit-`BackendConfig` and no-`BackendConfig` branches call:                                                                                                                                                                     
                                                                                                                                                                                                                                                                                       
  ```go                                                                                                                                                                                                                                                                                  
  getLabelsAsHCL(terraform.Labels, 6)                                                                                                                                                                                                                                                  
  ```                                                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                       
  rather than `terraform.Spec.BackendConfig.Labels` (or a merged view). `Spec.BackendConfig.Labels` is never read anywhere in the controller. Users setting it to `{}` expecting to strip user labels from the backend config experience no change in behavior — the kubernetes backend  
  still receives the full `metadata.labels` set.                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                         
  Once a CR is labelled (by Flux, ArgoCD, operators, or any mutating admission controller), every one of those labels flows into the kubernetes backend's selector, and any subsequent change to that label set breaks state reads.                                                      
                                                                                                                                                                                                                                                                                       
  ## Proposed fix                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                       
  In `controllers/tf_controller_backend.go`, resolve the backend HCL `labels` from `Spec.BackendConfig.Labels` when it is non-nil, falling back to `terraform.Labels` otherwise:                                                                                                         
                                                                                                                                                                                                                                                                                       
  ```go                                                                                                                                                                                                                                                                                  
  func backendHCLLabels(tf *infrav1.Terraform) map[string]string {                                                                                                                                                                                                                     
      if tf.Spec.BackendConfig != nil && tf.Spec.BackendConfig.Labels != nil {                                                                                                                                                                                                           
          return tf.Spec.BackendConfig.Labels                                                                                                                                                                                                                                            
      }                                                                                                                                                                                                                                                                                  
      return tf.Labels                                                                                                                                                                                                                                                                   
  }                                                                                                                                                                                                                                                                                      
  ```                                                                                                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                         
  Then replace both call sites with `getLabelsAsHCL(backendHCLLabels(terraform), 6)`.                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                         
  Design notes to decide in review:                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                       
  - **Backwards compatibility.** Today all users implicitly depend on `terraform.Labels` being propagated. Making `Spec.BackendConfig.Labels` *replace* rather than *merge* the metadata labels is a behavior change only for users who explicitly set that field. They can't be getting 
  anything useful from it today (it's ignored), so the upgrade should be safe. A merge strategy (`Spec.BackendConfig.Labels` takes precedence, other labels inherited from `terraform.Labels`) is also defensible if we think users rely on tool-added labels (Flux, ArgoCD, kustomize)
  being carried into the backend.                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                       
  - **Change the default for unset `Spec.BackendConfig.Labels`?** A stricter fix would restrict the default set to a minimal known-safe list (empty, or just the tofu-controller-managed ones) so that future label drift on the CR cannot wedge state. That is a bigger behavior change;
   doing both in sequence seems wise:                                                                                                                                                                                                                                                  
    - Honour `Spec.BackendConfig.Labels` now to give users a way out.                                                                                                                                                                                                                    
    - Consider flipping the default in a later major release.                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                         
  - **Documentation.** `docs/use-tf-controller/with-a-custom-backend.md` should document the selector-based state read behavior and recommend setting `Spec.BackendConfig.Labels: {}` unless there is a specific reason to want certain labels on the Secret. It should also warn that   
  **adding a label to a `Terraform` CR after its state Secret already exists will silently corrupt state persistence** unless the label is added to `Spec.BackendConfig.Labels` or the state Secret is deleted.                                                                          
                                                                                                                                                                                                                                                                                         
  - **Optional: defensive state-read behavior.** A future hardening would be to have tofu-controller watch its state Secrets and warn when the Secret's label set diverges from the configured HCL label set. Out of scope for this bug-fix PR.                                          
                                                                                                                                                                                                                                                                                       
  ### Test plan                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                       
  - Unit test `tf_controller_backend.go`: assert that when `Spec.BackendConfig.Labels` is set, the generated `backend_override.tf` contains exactly those labels; when it is nil, it falls back to `terraform.Labels` (status quo).                                                      
  - Integration test: create a `Terraform` CR with `spec.backendConfig.labels: {}`, apply, then add a custom label to `metadata.labels`, trigger another reconcile, verify `terraform plan` shows no changes (state was read correctly from the Secret).                               
  - Regression: verify existing test cases under `controllers/tc000020_*_with_backend_*` still pass.                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                         
  ## Workarounds (for users hitting this before the fix lands)                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                         
  1. Do not add or modify custom labels on the `Terraform` CR after its state Secret has been created, OR                                                                                                                                                                                
  2. Delete the state Secret so terraform re-creates it with the current full label set (only acceptable if losing state is acceptable), OR                                                                                                                                            
  3. Switch the backend to a non-kubernetes state backend (S3/GCS/etc) via `spec.backendConfig.customConfiguration` — the selector-based read logic is specific to the kubernetes backend.

matikij:
The kubernetes backend was always generating its label selector from terraform.Labels (the CR's metadata.labels). Adding a label to a CR after its state Secret had been created caused the backend selector to drift from the Secret's label set, yielding an empty List and a bootstrap wipe of state.

Use Spec.BackendConfig.Labels when non-nil (empty map is a valid explicit opt-out), otherwise fall back to terraform.Labels to preserve existing behavior.

Closes https://github.com/flux-iac/tofu-controller/issues/1774


matikij:
Fixed by #1808

PR #1826, opened 2026-06-16
chore!: remove unused option "--cert-validity-duration"

artem-nefedov:
This option has been unused for the long time. Other option "--ca-cert-validity-duration" is actually responsible for all certs duration, not just CA.

It was extra confusing that it's marked as "Deprecated", which normally means "it still works but you should not use it", while in reality it was completely ignored. The fact that helm chart still set it only added to the confusion.


artem-nefedov:
Something going on with dependencies in tests. Not related to changes in the PR.

alexandermarston (approved):
Thank you! LGTM 🚀

PR #1829, opened 2026-06-19
 Add libc6-compat for boundary

cschindlbeck:


cschindlbeck:
dup https://github.com/flux-iac/tofu-controller/pull/1825

PR #1835, opened 2026-07-02
fix: add controller ownerReference to tf-runner pods

### LINKED ISSUE #1832: tf-runner pod has no controller ownerReference → observability reports false "restart" alerts
### Description

The plan/apply runner pod (`<terraform-name>-tf-runner`) is created as a **bare pod with no `ownerReferences`** and a **stable name**, and with the default `alwaysCleanupRunnerPod: true` it is **created and deleted on every reconcile**.

Observability tooling that classifies pods by their controller ownerReference — kube-state-metrics' `kube_pod_info{created_by_kind}` and anything built on it (e.g. Coroot) — therefore treats the runner as a long-lived static pod whose single instance **"restarts"** on every reconcile. The result is a steady stream of false `instance restarted N times` alerts, even though the runner is perfectly healthy (the `Terraform` stays Ready / no drift) and the container `restartCount` is always `0` — it is pod *recreation*, not a container restart.

There's no clean in-cluster way to paper over it either: `runnerPodTemplate` exposes only `metadata.{labels,annotations}` (no `restartPolicy`, no `ownerReferences`), and forging an owner via a mutating webhook needs a fake placeholder Job/CronJob (GC requires the owner to exist) — a hack.

### Proposed solution

Set a controller `ownerReference` on the runner pod pointing to the `Terraform` CR that owns it (`controller: true`). That makes `created_by_kind` non-empty so observability classifies it as a managed pod rather than a restarting static pod, and as a bonus gives a clean GC cascade — runner pods are reaped when their `Terraform` CR is deleted, instead of relying solely on `alwaysCleanupRunnerPod`. (If maintainers prefer, running the runner as a `Job`-owned pod would additionally let tools that special-case Job/CronJob treat each recreation as an expected ephemeral task — happy to follow whichever direction you prefer.)

`alwaysCleanupRunnerPod: true` (the resource-efficient default) is desirable to keep — the gap is purely the missing ownerReference on the recreated runner pod.

I'd be happy to put up a PR for this.

### Environment

- tofu-controller / tf-runner `v0.16.4`
- Terraform CR `infra.contrib.fluxcd.io/v1alpha2`

devantler:
Fixes #1832

### Problem

The runner pod (`<terraform-name>-tf-runner`) is created as a bare pod with no `ownerReferences` and a stable name. With the default `alwaysCleanupRunnerPod: true` it is created and deleted on every reconcile, so observability tooling that classifies pods by their controller owner — kube-state-metrics' `kube_pod_info{created_by_kind}` and anything built on it (e.g. Coroot) — treats it as a long-lived static pod whose single instance "restarts" on every reconcile, producing a steady stream of false *instance restarted* alerts even though the runner is healthy and `restartCount` is always `0`.

### Solution

Set a controller `ownerReference` on the runner pod pointing to the `Terraform` object that owns it, mirroring the existing owner reference this project already sets on the outputs Secret (`runner/server_outputs.go`). The reference intentionally omits `blockOwnerDeletion`, matching that precedent, so no additional `terraforms/finalizers` permission paths are exercised on clusters running the `OwnerReferencesPermissionEnforcement` admission plugin.

Effects:

- `created_by_kind`/`created_by_name` become `Terraform`/`<name>`, so observability tools classify the runner as a managed pod instead of a restarting static pod.
- Kubernetes garbage collection reaps runner pods when their `Terraform` object is deleted — a safety net alongside `alwaysCleanupRunnerPod`. During a finalizer-driven `destroy`, the object still exists until the finalizer is removed, so GC cannot reap the runner pod mid-destroy.

The pod is always created in the same namespace as its `Terraform` object, so the owner reference is valid (no cross-namespace ownership).

### Testing

- Extended `Test_000260_runner_pod_test` to pin the owner reference (apiVersion, kind, name, UID, `controller: true`) on the generated pod template.
- `go build ./...`, `go vet ./controllers`, and the `Test_000260_runner_pod_test*` template tests pass locally against envtest (k8s 1.36.2).

> As disclosed in #1832, this change was planned and implemented with AI assistance (Claude), and reviewed by me.


devantler:
> 🤖 Generated by the Daily AI Assistant

Pushed a follow-up commit bumping `golang.org/x/crypto` v0.50.0 → v0.52.0 and `golang.org/x/net` v0.53.0 → v0.55.0 (both indirect) to clear the 13 HIGH Trivy findings that were failing the Scan workflow on this PR. `go mod tidy` + `go build ./...` pass locally; no source changes beyond the dependency bumps.

artem-nefedov:
I wonder what the current behavior is when you delete Terraform  with `destroyResourcesOnDeletion: false` while it has in-progress runner pod. I assume controller should only removes finalizer from Terraform after runner finishes, but it may be worth double-checking. If it does work like that, this change is ok.

devantler:
I checked this rather than assuming, and your assumption is correct.

The finalizer is only removed after the runner has finished. Two things make that hold:

- Deletion always uses the runner, even with `destroyResourcesOnDeletion: false`. The `FinalizeSecrets` call sits outside that flag's block, so it runs on every delete.
- Removing the finalizer is the last thing `finalize()` does. The object cannot be removed from the cluster while the finalizer is still on it, so garbage collection never gets the chance to take the runner pod early.

I confirmed the ordering with temporary log lines in `finalize()` and `reconcile()`, on both settings of the flag, under envtest:

```
calling finalize() destroyOnDeletion=false
FINALIZER REMOVED
reconcile() ENTERED deleting=true finalizers=0
setupTerraform RETURNED ERROR: Terraform.infra.contrib.fluxcd.io "..." not found
```

One thing there surprised me, so I want to flag it. `Reconcile` keeps going after `finalize()` succeeds, so a second `reconcile()` starts after the finalizer is already gone. It fails immediately, because the object no longer exists, and never gets as far as plan or apply. So nothing meaningful runs against the runner pod at that point, with or without this change.

The cause is at `tf_controller.go:525`:

`if terraform, result, err := r.finalize(...); err != nil {`

`:=` makes those three variables local to the if, so the only return is on the error path. That looks like a pre-existing bug. Happy to raise it separately if you want it tracked.

One limit of my testing, to be clear about it: envtest does not run kube-controller-manager, so I verified the ordering, not garbage collection timing itself.

alexandermarston (approved):

PR #1857, opened 2026-08-03
fix: do not reconcile after restart for healthy resources

### LINKED ISSUE #1681: Terraform is always reconciled after controller restart if plan is not updated and resource age > interval
Currently, the logic in `shouldReconcile()` compares current time to `status.lastPlanAt` value + interval duration [here](https://github.com/flux-iac/tofu-controller/blob/417a454df42c1bbe0240f8c887dcd0ad5a347039/controllers/tf_controller.go#L603) and [here](https://github.com/flux-iac/tofu-controller/blob/417a454df42c1bbe0240f8c887dcd0ad5a347039/controllers/tf_controller.go#L612). But the problem is that `status.lastPlanAt` isn't always updated on reconcile, e.g. if there are no planned changes it won't be updated.

So, what happens is after initial resource creation and planning, the logic will work correctly until time specified in `spec.interval` passes. After that, if there are no changes in plan, it is guaranteed to trigger reconcile because `status.lastPlanAt` + interval will always be lower than current time, regardless of when last reconcile happened.

Terraform resource also has `status.lastHandledReconcileAt`, which seems to be a better fit for comparison, as it will always be updated on a reconcile.

artem-nefedov:
fixes #1681

Actually makes controller not re-trigger reconciles after restart, assuming the resource is not failing.

It couldn't be done with existing status fields, so a new one had to be added.
The field currently only stores last successful run.

Tested, seems to work well.

Alternative is to do something like LastAttemptAt, which would store every run.
I haven't done it like that because retriggering reconcile for failed resources only didn't look that bad to me, and also there's still a guard with LastPlanAt right before that sometimes work.
LastSuccessfulReconcileAt also looked more useful from observability standpoint, e.g. with kube-state-metrics.
However, I'm open to opinions.


artem-nefedov:
test failure looks unrelated

alexandermarston (approved):
Nice improvement - thanks!

PR #1890, opened 2026-09-14
fix: apply the approved plan when the approval commit moves the source

### LINKED ISSUE #1878: Manual approval loops indefinitely when Terraform CR and source share a GitRepository
## Summary

Manual plan approval enters an infinite loop when the `Terraform` CR manifest and the Terraform configuration are stored in the same GitOps repository.

The manual approval documentation instructs users to copy the pending plan ID into `spec.approvePlan` and push the updated manifest to Git:

https://flux-iac.github.io/tofu-controller/use-tf-controller/plan-and-manually-apply-terraform-resources/

However, committing that approval changes the revision of the same `GitRepository` referenced by the Terraform resource. Because plan IDs are derived from the source revision, the controller generates a new plan ID before applying and rejects the approval.

This appears related to the pre-apply replan added in:

https://github.com/flux-iac/tofu-controller/pull/1640

Relevant code:

https://github.com/flux-iac/tofu-controller/blob/v0.16.5/controllers/tf_controller_reconcile.go#L201-L225

Plan ID generation:

https://github.com/flux-iac/tofu-controller/blob/v0.16.5/api/planid/plain_id.go

## Environment

- Tofu Controller: `v0.16.5`
- Flux-managed `Terraform` CR
- The `Terraform` manifest and Terraform configuration are stored in the same Git repository
- `sourceRef` points to the Flux `GitRepository` for that repository
- Manual approval mode

## Steps to reproduce

1. Store both the Terraform configuration and its `Terraform` CR in the same repository.
2. Push Terraform changes as commit `AAAA`.
3. Wait for the controller to produce:

   ```yaml
   status:
     plan:
       pending: plan-main-AAAA
   ```

4. Update the Git-managed manifest:

   ```yaml
   spec:
     approvePlan: plan-main-AAAA
   ```

5. Commit and push this approval as commit `BBBB`.

## Actual behavior

The approval commit changes the source revision from `AAAA` to `BBBB`.

During reconciliation, the controller runs another plan before manual apply. The new plan receives an ID based on the new revision:

```text
approvedPlan = plan-main-AAAA
pendingPlan  = plan-main-BBBB
```

The controller concludes that the plan changed and waits for another approval.

Approving `plan-main-BBBB` requires another commit, `CCCC`, which generates `plan-main-CCCC`. Repeating the documented process therefore creates an infinite loop.

A direct Kubernetes patch of `spec.approvePlan` works because it changes the Terraform CR without changing the referenced Git source revision.

## Logical issue

The pre-apply replan was added to ensure that the plan had not drifted while waiting for manual approval. However, the comparison is based on plan IDs derived from Git revisions rather than the contents of the plans.

Consequently, an approval-only Git commit produces a different plan ID even when the effective Terraform plan is unchanged.

This makes the documented Git-based manual approval flow incompatible with a common GitOps monorepo layout. The documentation does not state that the Terraform source and the Terraform CR must be stored in separate repositories.

## Expected behavior

Committing an approval for a specific pending plan should allow that exact saved plan to be applied.

An approval-only commit should not require an endless sequence of new approvals.

Any additional Terraform changes introduced after the approved plan should be planned and approved separately.

## Potential minimal fix

A possible minimal fix would be to restore saved-plan semantics for explicit manual approvals:

1. When `spec.approvePlan` matches `status.plan.pending`, do not generate and overwrite the plan before applying.
2. Load and apply the already saved plan corresponding to the approved ID.
3. Record `status.lastAppliedRevision` using `status.lastPlannedRevision`, since that is the revision from which the approved plan was created.
4. If the current source revision differs from the applied plan revision, immediately reconcile it afterwards:
   - If the effective Terraform configuration is unchanged, the result should be `No changes`.
   - If it contains additional Terraform changes, generate a new pending plan requiring separate approval.

This would ensure that only the plan explicitly reviewed by the user is applied, while still allowing the controller to reconcile a newer source revision afterwards.

An alternative, more extensive solution would be to compare normalized semantic plan digests rather than revision-derived plan IDs. Simply replacing the current ID with a constant or generic name would not be safe because an old approval could then approve a different future plan.

## Testing

A regression test could extend the existing manual approval tests with this sequence:

1. Generate a pending plan at source revision `AAAA`.
2. Set `spec.approvePlan` to that plan ID.
3. Simultaneously update the source revision to `BBBB`, simulating a Git approval commit without changing the effective Terraform configuration.
4. Verify that the saved plan from `AAAA` is applied.
5. Verify that `BBBB` is subsequently reconciled separately.

I would be happy to prepare a PR with the minimal fix and regression test if the maintainers agree with this behavior.

mloiseleur:
## Motivation

Manual approval loops forever when the `Terraform` object and the Terraform configuration share a Git repository.

The approval commit moves the source revision, and the pre-apply replan mints a plan ID that approval can never match (fixes #1878).


mloiseleur:
I'm pretty sure the failing test is not linked to my PR. 

@alexandermarston I opened https://github.com/flux-iac/tofu-controller/pull/1891 to fix the flaky test.

